PR #739, opened 2013-03-30
Allow plugins to be installed with shorthand user/repo notation

zeke:
This patch makes `heroku plugins:install tpope/heroku-binstubs` work just like `heroku plugins:install https://github.com/ddollar/heroku-redis-cli`

### DIFF
--- a/lib/heroku/plugin.rb
+++ b/lib/heroku/plugin.rb
@@ -91,6 +91,11 @@ module Heroku
     end
 
     def initialize(uri)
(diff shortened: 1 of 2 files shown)


mmcgrana:
Should we close this @dickeyxxx? Not sure what happened to the PR but I'm not sure it's relevant anymore.

dickeyxxx:
yeah, it's going to conflict with the new plugin architecture https://github.com/heroku/heroku-hello-world

Issue #256, opened 2012-03-14
stdin/stdout don't behave like UNIX with `heroku run`


ddollar:
Trying to run this command

```
cat /tmp/test.md | heroku run -a dx-dashboard "curl -s -F page=@- http://mantastic.herokuapp.com"
```

The pipes should all get hooked up correctly in a non-interactive mode to let stdin/stdout work properly with the shell container of heroku run.

Let me know if this doesn't make sense and I can try to elaborate.

PR #811, opened 2013-05-28
Send EOF to the remote on input EOF

dividedmind:
Attached patch sends 4.chr to the remote on input EOF. It finally makes it possible to pipe stuff to heroku run. Fixes issue #256

### DIFF
--- a/lib/heroku/client/rendezvous.rb
+++ b/lib/heroku/client/rendezvous.rb
@@ -56,6 +56,8 @@ class Heroku::Client::Rendezvous
             begin
               data = input.readpartial(10000)
             rescue EOFError
+              ssl_socket.write(4.chr)
+              ssl_socket.flush
               readables.delete(input)
               next
             end


pre:
Three years later, I have this same problem.:) I'm also trying to feed seed data to Rake from STDIN. 

Command:
`cat FILENAME | heroku run bash --noediting -c "rake db:seed:TASKNAME"`

Three notes:

1) If FILENAME has only a small amount of lines, reading from STDIN will always succeed, rake will receive EOF as expected and the process will stop eventually.

2) If FILENAME has 900 lines of plain text, reading from STDIN line-by-line `STDIN.each do |line|` will succeed, but rake will not receive EOF and the process will not stop.

3) If FILENAME has 900 lines of plaintext, reading from STDIN all-at-once `lines = STDIN.read` will echo contents on the screen, but rake will not receive EOF and will not process `lines`. Process will not stop.

```
heroku-toolbelt/3.43.9 (x86_64-darwin15) ruby/2.3.1
heroku-cli/5.2.39-010a227 (darwin-amd64) go1.6.2
=== Installed Plugins
heroku-certs@1.1.14
```

PR #813, opened 2013-06-13
heroku logs --tail doesnt work when the output is sent through a pipe

RedXVII:
```
heroku logs -t | cat
```

This command will not print any new output from the application, effectively ignoring the -t option.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -40,7 +40,7 @@ def prepare_command(klass)
   command
 end
 
-def execute(command_line)
+def execute(command_line, &block)
   extend RR::Adapters::RRMethods
 
   args = command_line.split(" ")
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -67,6 +67,8 @@ def tty?
     end
   end
 
+  yield $stdin, $stdout, $stderr if block_given?
+
   begin
     object.send(method)
   rescue SystemExit
```
sstabley8385: Change

rattrayalex:
FWIW, it does appear that `heroku --log -t > some_file.log` works, and you can `tail -f some_file.log | program` as a workaround

dickeyxxx:
I'm not able to reproduce this at all. Does `heroku logs -t | cat` work now? This code has been entirely rewritten.

PR #847, opened 2013-07-15
Enabled recursive Git submodule initialization for `heroku git:clone`

wilkinson:
This is a pretty simple edit that allows convenient use of Git submodules from within the Heroku client. All it does is add the "--recursive" flag to the system call to Git; this will automatically initialize and update [potentially nested] submodules. This can be very useful for collaborating on large, complex projects! ([documentation](http://git-scm.com/docs/git-submodule))

### DIFF
--- a/lib/heroku/command/git.rb
+++ b/lib/heroku/command/git.rb
@@ -28,7 +28,7 @@ class Heroku::Command::Git < Heroku::Command::Base
     git_url = api.get_app(name).body["git_url"]
 
     puts "Cloning from app '#{name}'..."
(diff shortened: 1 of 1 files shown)


Enelar:
@wilkinson could you please continue your attemts?
I suffer without this feature

dickeyxxx:
Just add --recursive. It will pass through

Enelar:
@dickeyxxx where? I use GitHub integration, and there no way to add arguments

dickeyxxx:
This PR has nothing to do with github integration

PR #857, opened 2013-07-27
hide passwords by default

kr:
Hey folks, here's a stop gap patch to address the problem of accidentally
leaking passwords. It certainly isn't a full solution, but it might be helpful.

What do you think?

I'd appreciate pointers on how to write unit tests for this change.
The readme isn't too helpful:

```
$ grep -iA2 test README.md 
If you're working on the CLI and you can smoke-test your changes:

    $ bundle exec heroku
```

Who looks at this repo these days?

cc @fdr

### DIFF
--- a/lib/heroku/command/config.rb
+++ b/lib/heroku/command/config.rb
@@ -1,3 +1,4 @@
+require "uri"
 require "heroku/command/base"
 
 # manage app config vars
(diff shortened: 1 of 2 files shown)


kr:
I bet this is stale enough to be irrelevant. (Most of the people on this PR don't even work at Heroku any more.) Closing.

PR #905, opened 2013-09-21
interactively execute multiple commands

wh0:
On my computer, it takes a long time to boot Ruby and do all the `require`s.
So I made this interactive loop thing that lets you run multiple commands.


trimble:
This might be an interesting use case, but right now all the tests fail. Is this still worth pursuing? Or should this be closed?

wh0:
I no longer have the problem that this change solves.

PR #1346, opened 2015-01-05
Add certs:generate command

brentdax:
This pull request adds a `certs:generate` command to the Heroku client.

Generating an SSL certificate involves long, opaque OpenSSL incantations. The Web is full of them, each slightly different from the others. Some generate inadequately small keys. Most password-protect the key file, which the Heroku client doesn't like. All of them are easy to mistype or mis-paste and impossible to remember.

`certs:generate` creates a CSR and adequately-sized key in one easy to remember step. It then instructs the user on the next command to use (`certs:add` or `certs:update`, depending on whether the domain has already been added to an SSL Endpoint). I think this is a much better approach than the status quo.

**This pull request is not yet ready to merge.** Among its issues:

* I haven't written any tests. (I'm not even sure how you *would* test this, honestly.)
* I've only tried using this on OS X Yosemite. I'm pretty sure it'll work fine on Linux, but Windows might be trouble—it shells out to openssl(1) to do the hard work, so we may need to test if OpenSSL is installed and tell Windows users to install it if it's not.
* The "UI" isn't really where I'd like it to be yet. When you don't specify a subject for the certificate, OpenSSL's interactive mode doesn't default to using the specified domain as the common name, which I'd really like to see; when you *do* specify a subject, you must do so in a very opaque format. I'd love to see ideas for improvement here.
* Some people might prefer a larger key than 2048 bits. This is the size created by [the commands in Heroku's documentation](https://devcenter.heroku.com/articles/ssl-endpoint#generate-private-key), but we can argue about whether it should be bigger, or if there should be an option to control its size.

If this is something you'd be interested in merging when complete, please let me know and I'll see about finishing it. (Or you can take over—I'm not attached to it.)


dickeyxxx:
this looks great @brentdax! one idea @brandur had was for a `--self-signed` flag. I think it would be good to check to see if `openssl(1)` is available, but assuming that is in place and the test suite is fixed I'd be happy to merge.

dickeyxxx:
looks great! thanks!

brandur:
Thanks @brentdax! Great work.

PR #1348, opened 2015-01-05
show warning if HTTPS_PROXY is set and there is a certificate error

jdx:
@geemus curious if you have any thoughts on this


dickeyxxx:
actually this is no good since it's just under that logs method... looking for a better place to add this

PR #1351, opened 2015-01-10
Revert "Add certs:generate command"

dickeyxxx:
Reverts heroku/heroku#1346

@brentdax looks like those tests are a bit flaky https://travis-ci.org/heroku/heroku/jobs/46520477


brentdax:
Oof. `ensure_openssl_installed!` includes some code to make sure it only actually performs its test once; this wasn't always cleared between tests, so running the tests in the wrong order would cause the test that confirms the method actually calls out to `openssl(1)` to fail.

I've pushed a quick patch to my branch at dd5e7f64 which fixes this issue.

dickeyxxx:
You'll need to make a new PR

brentdax:
@dickeyxxx Done (after a false start) as #1353. Thanks!

PR #1353, opened 2015-01-12
feature/certs-generate (this time for sure!)

brentdax:
Reverts the reversion of my previous pull request to add a `certs:generate` command, and integrates a fix for the flaky `ensure_openssl_installed!` test that caused tests to sometimes fail.

### DIFF
--- a/lib/heroku/command/certs.rb
+++ b/lib/heroku/command/certs.rb
@@ -1,4 +1,5 @@
 require "heroku/command/base"
+require "heroku/open_ssl"
 require "excon"
 
 # manage ssl endpoints for an app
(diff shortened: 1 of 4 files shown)


dickeyxxx:
looks like there is still an issue https://travis-ci.org/heroku/heroku/jobs/46705089

brentdax:
Yikes, looks like this hit a rare corner case. (The test fails if it's the first in its group, but comes after another group that implicitly calls the method it's testing.) To fix the issue, I made the resetting much more aggressive. I can't be totally sure that the issue won't recur, but I arranged for my computer to run `spec/heroku/open_ssl_spec.rb` 1,000 times and the full suite 100 times and haven't seen any failures, so I think it's finally nailed to the wall.

dickeyxxx:
lgtm!

PR #1354, opened 2015-01-12
Fix bug in extract_app_in_dir when getting remote

deeperx:
Sometimes, heroku client cannot find the remote app even when it's been added to git config. This fixes it.

**Case:**

```bash
git remote add heroku git@heroku.com:yourapp.git

# OR
heroku git:remote -a yourapp
```

```bash
# .git/config
[remote "heroku"]
	url = https://git.heroku.com/yourapp.git
	fetch = +refs/heads/*:refs/remotes/heroku/*
```

```bash
$ heroku logs
!    No app specified.
!    Run this command from an app folder or specify which app to use with --app APP.
```

### DIFF
--- a/lib/heroku/command/base.rb
+++ b/lib/heroku/command/base.rb
@@ -218,7 +218,7 @@ protected
     if remote = options[:remote]
       remotes[remote]
     elsif remote = extract_app_from_git_config
-      remotes[remote]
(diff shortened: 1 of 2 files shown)


dickeyxxx:
closing since this was resolved in #1357

PR #1356, opened 2015-01-13
ensured newline is added before additions to startup file

dickeyxxx:


dickeyxxx:
fixes https://github.com/heroku/toolbelt/issues/121

PR #1367, opened 2015-01-20
Use Cedar Stack in `heroku create` Help

naaman:
This updates the `heroku create` help to show a cedar stack example
instead of the [deprecated bamboo stack](https://devcenter.heroku.com/articles/bamboo).


catsby:
How about `cedar-14`?

naaman:
Cedar 14 is the default, so that'd be kinda weird IMO.

catsby:
Maybe.. there's no mention in the help text what the default is though. As is, I wonder if viewing the help text here actually suggests people make `cedar` apps. 

Adding mention of the default on line 196 and leaving as you have it:

```
# -s, --stack STACK          # the stack on which to create the app. Default is cedar-14"
```

as well as updating the current examples to read `Creating <app>...done, stack is cedar-14` would clarify things for me.

This is nitpicky though :smile:

PR #1372, opened 2015-01-26
Move pg:backups commands from pg-extras into main CLI

uhoh-itsmaciek:


uhoh-itsmaciek:
Looking into the CI failures.

PR #1381, opened 2015-01-27
Revert "Allow use of remote source if URI is specified as database."

dickeyxxx:
Reverts heroku/heroku#1369

We got some bug complaints with the following error:

```
 !    Remote database is not empty.
 !    Please create a new database, or use `heroku pg:reset`
```

cc @tmertens

### DIFF
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -306,14 +306,10 @@ class Heroku::Command::Pg < Heroku::Command::Base
   end
 
 
-  # pg:push <SOURCE_DATABASE> <REMOTE_TARGET_DATABASE>
+  # pg:push <LOCAL_SOURCE_DATABASE> <REMOTE_TARGET_DATABASE>
   #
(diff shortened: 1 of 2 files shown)


tmertens:
@dickeyxxx Can you provide more information?  Isn't the above error the expected behavior of pg push is used to push data to a heroku db that is not empty?

https://github.com/heroku/heroku/blob/master/lib/heroku/command/pg.rb#L312

It doesn't sound related to my PR which is no-op for the case of local databases.

tmertens:
actually i may see a bug in my patch... better tests for the pg commands would be nice but there wasnt much to start with. I will take a look and see about fixing it.

dickeyxxx:
I didn't investigate but we got a lot of users complaining about this feature right away. @recurrence can you enlighten us on what the specific problem was?

tmertens:
Yeah ... I see the bug. Sorry about that.  Easy fix but i'm gonna spend some time seeing if I can easily add some tests for the pg push/pull commands to prevent refactoring issues in the future.

dickeyxxx:
no problem, I should've tested it myself first

PR #1383, opened 2015-01-28
Show a better error message for git failures.

dickeyxxx:
If git is not installed or isn't outputting the version from `git --version`
then we show this message.


friism:
I think that optimally, we shouldn't point people at GitHub directly. How about http://git-scm.com/ ?

dickeyxxx:
yeah I suppose that's what the github link links to anyways

PR #1392, opened 2015-01-29
prevent fork from deleting existing apps

dickeyxxx:


geemus:
:+1: awesome, glad to see a simple/quick fix for this edge case.

PR #1395, opened 2015-01-30
remove default org functionality in favor of HEROKU_ORGANIZATION

dickeyxxx:


dickeyxxx:
one thing I still need to do is keep the org on fork

raulb:
@dickeyxxx omg +:100:

PR #1396, opened 2015-01-31
warn when using jruby with heroku run commands

dickeyxxx:
to help with https://github.com/heroku/heroku-buildpack-ruby/issues/178


dickeyxxx:
cc @schneems

schneems:
Awesome sauce! Thanks :heart:

PR #1398, opened 2015-02-02
Updated help text for 2fa in two_factor.rb

mbren:
Edited the CLI help text for the twofactor command and moved the first comment directly before the definition of the TwoFactor class to resolve the issue that the main help text for the twofactor command was missing:
  ```
  stack           #  manage the stack for an app
  status          #  check status of heroku platform
  twofactor       #                   
  update          #  update the heroku client
  version         #  display version
```

I did not test this fix so please double check that everything is ok before you merge it.

### DIFF
--- a/lib/heroku/command/two_factor.rb
+++ b/lib/heroku/command/two_factor.rb
@@ -1,12 +1,12 @@
 require "heroku/command/base"
 
-# manage two factor settings for account
-#
 module Heroku::Command
+  # manage two-factor authentication settings for your account
+  #
   class TwoFactor < BaseWithApp
     # 2fa
     #
-    # Display whether two-factor is enabled or not
+    # Display whether two-factor authentication is enabled or not
     #
     def index
       account = api.request(
@@ -16,9 +16,9 @@ module Heroku::Command
         :path    => "/account").body
 
       if account["two_factor_authentication"]
-        display "Two-factor auth is enabled."
+        display "Two-factor authentication is enabled."
       else
-        display "Two-factor is not enabled."
+        display "Two-factor authentication is not enabled."
       end
     end
 
@@ -26,7 +26,7 @@ module Heroku::Command
 
     # 2fa:disable
     #
-    # Disable 2fa on your account
+    # Disable two-factor authentication for your account
     #
     def disable
       print "Password (typing will be hidden): "
@@ -52,7 +52,7 @@ module Heroku::Command
 
     # 2fa:generate-recovery-codes
     #
-    # Generates (and replaces) recovery codes
+    # Generates and replaces recovery codes
     #
     def generate_recovery_codes
       code = Heroku::Auth.ask_for_second_factor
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by dickeyxxx** [comment]
thanks! This seems wrong still though (although nothing that you changed)

```
☸ ~/s/g/h/heroku heroku help twofactor
Usage: heroku 2fa

 Display whether two-factor authentication is enabled or not

Additional commands, type "heroku help COMMAND" for more details:

  2fa:disable                  #  Disable two-factor authentication for your account
  2fa:generate-recovery-codes  #  Generates and replaces recovery codes
```

Should we change the help text from `2fa` to `twofactor`?

**Comment by dickeyxxx** [comment]
Also, I think we should stick with just `auth` instead of `authentication` since now it's our longest message

```
☸ ~/s/g/h/heroku heroku help
Usage: heroku COMMAND [--app APP] [command-specific-options]

Primary help topics, type "heroku help TOPIC" for more details:

  addons    #  manage addon resources
  apps      #  manage apps (create, destroy)
  auth      #  authentication (login, logout)
  config    #  manage app config vars
  domains   #  manage custom domains
  logs      #  display logs for an app
  ps        #  manage dynos (dynos, workers)
  releases  #  manage app releases
  run       #  run one-off commands (console, rake)
  sharing   #  manage collaborators on an app

Additional topics:

  certs        #  manage ssl endpoints for an app
  commands     #  list all commands
  drains       #  display drains for an app
  features     #  manage optional features
  fork         #  clone an existing app
  git          #  manage git for apps
  help         #  list commands and display help
  keys         #  manage authentication keys
  labs         #  manage optional features
  maintenance  #  manage maintenance mode for an app
  members      #  manage membership in organization accounts
  orgs         #  manage organization accounts
  pg           #  manage heroku-postgresql databases
  pgbackups    #  manage backups of heroku postgresql databases
  plugins      #  manage plugins to the heroku gem
  regions      #  list available regions
  stack        #  manage the stack for an app
  status       #  check status of heroku platform
  twofactor    #  manage two-factor authentication settings for your account
  update       #  update the heroku client
  version      #  display version
```

dickeyxxx:
in fact, I sort of think it should just be `manage two-factor settings`. It's clear that it has to do with authentication and `for your account` is obvious.

mbren:
"Two-factor" is an adjective that qualifies "authentication", the noun. I strongly feel that the term we use should be "two-factor authentication" or "two-factor auth". If we just say two-factor by itself, then we are using two-factor as a noun which, in my opinion, is not correct. 

However, at the end of the day, what matters is that there's useful help text for people to read. It's not like we're localizing this and have to worry about being consistent for a translation memory database. Do whatever you think is best.

I'm OK with the rest of your suggestions.

dickeyxxx:
could we make it `manage two-factor authentication settings`?

mbren:
yes

mbren:
:sparkles: 
Thank you!

jonmountjoy:
:+1: - thanks for spotting that this was missing @mbren

PR #1414, opened 2015-02-09
switched regex to match pg:backups unschedule color

### LINKED ISSUE #1413: Can't pg:backups unschedule
It looks like some naming conventions have moved around:

```
$ heroku pg:backups schedules
=== Backup Schedules
HEROKU_POSTGRESQL_VIOLET_URL: daily at 7:00 (UTC)

$ heroku pg:backups unschedule VIOLET
No automatic daily backups for HEROKU_POSTGRESQL_VIOLET found
```

It's related to a regex match. I'll open a PR shortly with a fix.

chadbailey59:
Fixes #1413 . @uhoh-itsmaciek , Can you double-check that I'm using color names correctly here?

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -397,7 +397,9 @@ EOF
     attachment = generate_resolver.resolve(db, "DATABASE_URL")
 
     schedule = hpg_client(attachment).schedules.find do |s|
-      attachment.name =~ /#{s[:name]}/
+      # attachment.name is HEROKU_POSTGRESQL_COLOR
+      # s[:name] is HEROKU_POSTGRESQL_COLOR_URL
+      /#{attachment.name}/ =~ s[:name]
     end
 
     if schedule.nil?
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -399,7 +399,7 @@ EOF
     schedule = hpg_client(attachment).schedules.find do |s|
       # attachment.name is HEROKU_POSTGRESQL_COLOR
       # s[:name] is HEROKU_POSTGRESQL_COLOR_URL
-      /#{attachment.name}/ =~ s[:name]
+      s[:name] =~ /#{attachment.name}/ || attachment.name =~ /#{s[:name]}/
     end
 
     if schedule.nil?


lib/heroku/command/pg_backups.rb
```diff
@@ -397,7 +397,9 @@ def unschedule_backups
     attachment = generate_resolver.resolve(db, "DATABASE_URL")
 
     schedule = hpg_client(attachment).schedules.find do |s|
-      attachment.name =~ /#{s[:name]}/
+      # attachment.name is HEROKU_POSTGRESQL_COLOR
+      # s[:name] is HEROKU_POSTGRESQL_COLOR_URL
+      /#{attachment.name}/ =~ s[:name]
```
uhoh-itsmaciek: Can you reverse this check to `s[:name] =~ /#{attachment.name}/`? I think that's semantically the same, but more idiomatic (I've never seen a regex on the left of `=~` although it does seem to work).

uhoh-itsmaciek: Actually, maybe let's include both checks with `||`--that will make sure we handle older schedules as well.

chadbailey59:
@uhoh-itsmaciek I added what I think you're asking for, although I can't test it, because I seem to have put my test app in a wierd unknown state. Also, I didn't see a spec at `spec/heroku/command/pg_backups_spec.rb`, so I didn't add tests. :postal_horn: SO BEWARE :postal_horn:

uhoh-itsmaciek:
Looks good. And yeah, unfortunately I picked up pg-extras' bad habit of lack of test coverage. I'll try to add some basic coverage when I have a chance.

chadbailey59:
Seems to be related to https://support.heroku.com/tickets/200070 . Who do I talk to about merging, etc?

dickeyxxx:
if @uhoh-itsmaciek thinks it looks good I'm comfortable with merging

uhoh-itsmaciek:
@neovintage you had run into this issue, right? Can you try @chadbailey59's fix and see if that addresses it?

You should be able to install and run a local toolbelt by running `git clone git@github.com:heroku/heroku.git` and then specifying the full path to the binary instead of relying on `$PATH` (`bin/heroku pg:backups unschedule`). Let me know if you run into any problems.

neovintage:
Awesome.  This pr fixes the problem of unscheduling.  

I found another problem. If you try to schedule over an already present schedule then we get an internal server error:  http://cl.ly/image/091Y1F1r291u.  We can open that as a bug and do it later.  Its not as critical because there is a workaround path (e.g. unschedule first then schedule).

uhoh-itsmaciek:
@neovintage That happens with or without this patch, right? I'll take a look at that, but if so, then I think this is good to go @dickeyxxx (though I'll have some more patches for you later today).

neovintage:
That happens with ***AND*** without this patch

uhoh-itsmaciek:
Ok, great. So I'm looking at that now, but I think this patch is ready to go.

PR #1422, opened 2015-02-24
Fixed OS checking on Arch Linux using JRuby

zcid:
Changed the operating system test to use a more relevant source. When using JRuby,
RAILS_PLATFORM will return "java" which is not relevant to the test.

Switched to RbConfig, a more reliable method for this.

### DIFF
--- a/lib/heroku/jsplugin.rb
+++ b/lib/heroku/jsplugin.rb
@@ -108,7 +108,7 @@ class Heroku::JSPlugin
   end
 
   def self.os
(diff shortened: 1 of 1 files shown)


dickeyxxx:
unfortunately this has it's own problem, it won't work under ruby 1.8.7. Perhaps change it to check if RbConfig exists then use RUBY_PLATFORM as a fallback?

dickeyxxx:
closing in favor of #1451

PR #1425, opened 2015-02-26
Better formatting for orgs:default message

dickeyxxx:


raulb:
For the record. It currently is:

```
$ heroku orgs:default personal
DEPRECATED: Use HEROKU_ORGANIZATION environment variable.
```

And this PR proposes:

```
$ heroku orgs:default
!    orgs:default is no longer in the CLI.
!    Use the HEROKU_ORGANIZATION environment variable instead.
!    See https://devcenter.heroku.com/articles/develop-orgs#default-org for more info.
```

@dickeyxxx Can we get rid of the reference of the [Dev Center article](https://devcenter.heroku.com/articles/1931) for now? Looking at it I realized there are too many bits that are currently wrong and we should fix. I'm not sure

Unless @itchymutt or @jsimone think we could easily revisit the whole article and make it right.

raulb:
@dickeyxxx Dev Center article is up to date now: https://devcenter.heroku.com/articles/1931#default-org.

dickeyxxx:
great!

PR #1429, opened 2015-02-27
Always Allow Org Opts

naaman:
Orgs are no longer a separate service and should work in alternate
clouds. This removes the `skip_org?` check in the command base.


dickeyxxx:
lgtm

PR #1435, opened 2015-03-02
Fix --wait-interval pg:wait flag

### LINKED ISSUE #1434: --wait-interval flag causes errors
`Error:       can't convert String into time interval (TypeError)`

It's trying to call `sleep` without converting the param to an integer.

JacobEvelyn:
Fixes #1434.


dickeyxxx:
this would leave it as 0 if the wait interval wasn't specified. Closing in favor of #1436

JacobEvelyn:
Ah right; I forgot about the `nil` case. Good catch! :)

PR #1437, opened 2015-03-03
Initial sketch for presenting cedar as cedar-10

cyx:
We discussed various ways to do this, and it seems
the least obtrusive way is to do this at the boundaries
(i.e. dashboard / cli) as opposed to changing internals.


dickeyxxx:
lgtm

PR #1446, opened 2015-03-11
Fork pgbackups

neovintage:
Transfer command fixed.  All databases copy correctly. 

/cc @dickeyxxx

### DIFF
--- a/lib/heroku/command/fork.rb
+++ b/lib/heroku/command/fork.rb
@@ -134,8 +134,8 @@ module Heroku::Command
         to_config = api.get_config_vars(to).body
         to_attachment = to_addon["message"].match(/Attached as (\w+)_URL\n/)[1]
 
-        resolver = Heroku::Helpers::HerokuPostgresql::Resolver.new(from, api)
-        attachment = resolver.resolve("#{from_attachment}_URL", nil)
+        resolver = Heroku::Helpers::HerokuPostgresql::Resolver.new(to, api)
+        attachment = resolver.resolve("#{to_attachment}_URL", nil)
         pgb = Heroku::Client::HerokuPostgresql.new(attachment)
         transfer = pgb.pg_copy(
           from_attachment.gsub('HEROKU_POSTGRESQL_',''),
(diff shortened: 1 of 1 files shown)


uhoh-itsmaciek:
[This](https://github.com/heroku/heroku/blob/fork-pgbackups/lib/heroku/command/fork.rb#L130-L136) should go, along with its call sites.

lib/heroku/command/fork.rb
```diff
@@ -144,20 +144,22 @@ def migrate_db(from_addon, from, to_addon, to)
...
         to_attachment = to_addon["message"].match(/Attached as (\w+)_URL\n/)[1]
 
-        pgb = Heroku::Client::Pgbackups.new(from_config["PGBACKUPS_URL"])
-        transfer = pgb.create_transfer(
+        resolver = Heroku::Helpers::HerokuPostgresql::Resolver.new(to, api)
+        attachment = resolver.resolve("#{to_attachment}_URL", nil)
+        pgb = Heroku::Client::HerokuPostgresql.new(attachment)
+        transfer = pgb.pg_copy(
+          from_attachment.gsub('HEROKU_POSTGRESQL_',''),
           from_config["#{from_attachment}_URL"],
```
uhoh-itsmaciek: So I'm not sure about the soundness of getting the URL like this, but you're not changing that aspect of this functionality, so I think it's fine for now--it's just something we may want to keep in mind, especially with the shareable add-on changes.

uhoh-itsmaciek:
+1, but it looks like you already merged, @dickeyxxx?

dickeyxxx:
yeah, I tested it and it seemed to work. I do agree with you the way it gathers the URL isn't ideal though

dickeyxxx:
but as you pointed out, it was like that before

neovintage:
I agree.  This was meant as more of stop-gap to get pgbackups using
transferatu.

uhoh-itsmaciek:
Ok, great. Thanks.

PR #1450, opened 2015-03-17
Add Request ID logging to STDERR

shirshendu:
Fixes #1212 
It's an old issue, but it does seem useful to have for CLI support requests (though I'm not the best person to comment on that).
@sclasen @dickeyxxx could you please have a look?

PR #1455, opened 2015-03-18
only show request id on an error

dickeyxxx:


lib/heroku/helpers.rb
```diff
@@ -283,7 +283,7 @@ def error(message, report=false, opts={})
       rollbar_id = Rollbar.error(message) if report
       $stderr.puts("Error ID: #{rollbar_id}") if rollbar_id
       request_id = opts[:error] && opts[:error].response ? opts[:error].response.headers['Request-Id'] : nil
-      $stderr.puts "Request ID: #{request_id}" if request_id
+      $stderr.puts "Request ID: #{request_id}" if report && request_id
```
shirshendu: @dickeyxxx Apologies for not seeing this earlier; I would like to understand this change a bit better.

This would imply that the `request_id` is only surfaced when `report` is true, which is never because in lib/heroku/command.rb
https://github.com/heroku/heroku/pull/1450/files#diff-b8c10b1653b4030aef2eada476d9a22aR232
report is being passed in as nil (which was the case before my PR as well, when it was not passed in at all).

I originally made this change for only non-rollbar error method calls. I'm having difficulty in finding the error method calls which have `report` true. Could you please guide me to them, because the error object needs to be passed in for them, else this request ID is never printed...

dickeyxxx: ah you're right. This was way too noisy, we often have the API respond with errors about things like an app name already existing. It's too noisy to have the request id on all of that. I'm going to roll back both changes.

PR #1470, opened 2015-03-23
Bump rest-client dependency to 1.7.3

Jberlinsky:
The currently pinned version of the `rest-client` gem, 1.6.7, has an active CVE: http://www.osvdb.org/show/osvdb/117461. While this is merely an information disclosure vulnerability, it breaks the build pipeline of many people who use something like `bundler-audit`, and should really be fixed anyways.

This pull request upgrades `rest-client` to the first non-vulnerable version, `1.7.3`.


Jberlinsky:
The tests fail to pass because of Ruby 1.8.7, which is no longer supported by `rest-client` as it no longer receives security patches.

mikeatlas:
:+1: please

mikeatlas:
Gemcanary is another service that detects and warns about this dependency's CVE. It's what brought me to come knocking on this gem's door and find this OPs thread. https://gemcanary.com/

flori:
Better make it rest-client >= 1.8.0 b/c there's CVE-2015-1820 now.

Also: It would be better not to depend so strictly on a specific point release of rest-client. I cannot perform a security upgrade for rest-client in one of my apps because heroku demands a very specific point release.

mikeatlas:
Bump. Please fix ASAP; we intentionally break our builds internally based off of failing `bundler-audit` checks for good reason!

cepko33:
:+1:

dickeyxxx:
We are discussing this one a bit internally as it will break 1.8.7 compatibility. I'm likely to merge and push it tomorrow along with #1479.

In any case, you shouldn't include the heroku gem in your gemfile, you should use the toolbelt and you won't be burdened with the gem versions we require.

mikeatlas:
Thanks @dickeyxxx, we don't reference the gem directly, but one we do use does [pgbackups-archive](https://github.com/kjohnston/pgbackups-archive/blob/master/) in our case. I'll flag an issue with that gem too, perhaps it can wrap CLI commands instead.

kjohnston:
@dickeyxxx, is/could there be any way of talking directly to services the Toolbelt wraps, namely the new Heroku PGBackups service?  This would allow gems like [pgbackups-archive](https://github.com/kjohnston/pgbackups-archive) to stop piggybacking onto the `heroku` gem or shelling out to the Toolbelt like [heroku-database-backups](https://github.com/kbaum/heroku-database-backups) does.

dickeyxxx:
the best way to interact with the API from ruby is to use this gem https://github.com/heroku/platform-api the gem the toolbelt uses is deprecated and uses the old v2 api. Unfortunately I don't know if the new pgbackups stuff is in there (or will be), can you enlighten us on that @uhoh-itsmaciek?

dickeyxxx:
if you're curious, the CLI is being replaced with a go/io.js implementation https://github.com/heroku/heroku-cli
That's why there aren't plans to replace the internals of the gem to use the platform api gem

dickeyxxx:
@Jberlinsky @kjohnston @mikeatlas @cepko33 update: I've got a list of a few users known to be still using 1.8.7. I'm reaching out to some of them to see why and if I can update them before I push out #1479

kjohnston:
@dickeyxxx, thanks for looking into the CLI/PGBackups access.

I know that you guys would love to deprecate this old gem and would rather invest the effort into the new CLI and Platform API gems.  The Platform API is really what those of us maintaining gems around areas like this crave, however I check in on it every now and then and it seems development has stalled and the documentation doesn't mention PGBackups.

If you hear anything further, please let me know.  @uhoh-itsmaciek, if you can share anything, please do, or let me know the proper channel.

uhoh-itsmaciek:
@kjohnston @dickeyxxx unfortunately, the backups functionality is not a part of the `platform-api` gem as it talks to a different system. We do not have a documented API or a standalone client for the functionality at this time. My colleague @neovintage may be able to tell you more.

kjohnston:
@uhoh-itsmaciek, thank you for confirming.  @neovintage, how might one be able to access the PGBackups API without using this deprecated gem?

I don't mean to hijack the thread here, it's just that these are the types of reasons that this gem needs to have ongoing maintenance discussed here.

neovintage:
@kjohnston The best advice I can give you right now is that you'll have to parse out the endpoints to the pgbackups API in a couple of files:
https://github.com/heroku/heroku/blob/master/lib/heroku/client/heroku_postgresql.rb
https://github.com/heroku/heroku/blob/master/lib/heroku/client/heroku_postgresql_backups.rb

An example of how it's used is in the fork functionality:
https://github.com/heroku/heroku/blob/master/lib/heroku/command/fork.rb#L131-L155

I recognize that's not the best answer and the more appropriate path would be to have officially supported endpoints for postgres and pgbackups.  It's a project I'd like to take on but probably not for some time.

kjohnston:
@neovintage, thank you for identifying these areas.

dickeyxxx:
still waiting for #1479 which I'm targeting for Wednesday. #1482 went out today to warn 1.8.7 users

PR #1476, opened 2015-03-31
Update rubyzip dependency.

### LINKED ISSUE #1474: Fixed runtime decency `rubyzip = 0.9.9` leds to incompatibility
I want to add [pgbackups-archive (1.0.0)](https://rubygems.org/gems/pgbackups-archive/versions/1.0.0) to my project. bundler can't finished due to Heroku requires a fixed `rubyzip = 0.9.9`

```
Bundler could not find compatible versions for gem "rubyzip":
  In Gemfile:
    pgbackups-archive (= 1.0.0) ruby depends on
      heroku (>= 3.28.6, ~> 3.28) ruby depends on
        rubyzip (= 0.9.9) ruby

    selenium-webdriver (= 2.45.0) ruby depends on
      rubyzip (1.0.0)
```

voxik:
Use more recent version or rubyzip. This will fix issues with old rubyzip on Fedora as well as it should resolve #1474

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -7,7 +7,7 @@ PATH
       multi_json (~> 1.10)
       netrc (>= 0.10.0)
       rest-client (= 1.6.7)
-      rubyzip (= 0.9.9)
+      rubyzip (~> 1.0)
 
 GEM
   remote: https://rubygems.org/
@@ -56,7 +56,7 @@ GEM
       diff-lcs (>= 1.2.0, < 2.0)
       rspec-support (~> 3.2.0)
     rspec-support (3.2.2)
-    rubyzip (0.9.9)
+    rubyzip (1.1.7)
     safe_yaml (1.0.4)
     simplecov (0.9.2)
       docile (~> 1.1.0)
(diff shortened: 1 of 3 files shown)


voxik:
The tests are failing due to Gemfile.lock. I didn't want to fiddle with that, since I don't think Gemfile.lock belongs into SCM.

voxik:
Another possibility is to go with zip-zip, but I don't think this is good idea ;)

dickeyxxx:
you're going to need to update the `Gemfile.lock` then. This is also going to break 1.8.7 compatibility, but that time has probably come.

dickeyxxx:
in gems you don't want to check in the lock file, but that's really not how the toolbelt is intended to be used. For the toolbelt, we need to lock down the dependencies to specific version numbers.

voxik:
Ok then. I forced the commit and updated the Gemfile.lock to rubyzip 1.1.7. I hope that is OK. The remaining failing test is the Ruby 1.8.7 as you already pointed out.

dickeyxxx:
still waiting for #1479 which I'm targeting for Wednesday. #1482 went out today to warn 1.8.7 users

dickeyxxx:
apologies @voxik but I wasn't able to merge this one due to some conflicts, but I did add the changes you made inside #1491

voxik:
Np. Thank you!

PR #1479, opened 2015-04-01
removed CLI support for ruby 1.8.7

dickeyxxx:
this is to be compatible with #1476 and #1470. It will be released 4/8.

This removes the ability to use the Heroku CLI from Ruby 1.8.7. **It does not affect apps running inside Heroku**, only the machine you use the CLI from.

Updating Instructions:

**Windows**
Simply reinstall the toolbelt from toolbelt.heroku.com

**OSX pkg**
Simply reinstall the toolbelt from toolbelt.heroku.com

**OSX Homebrew**
Do one of the following:
* Install a newer version of Ruby with `brew install ruby`
* Set your version of Ruby to something newer with rbenv or rvm
* Upgrade your version of OSX to Lion or above

**Linux**
On Linux you'll need to use your package manager to install a newer version of ruby. It will use whatever `ruby` is currently in the path.


lib/heroku/cli.rb
```diff
@@ -1,3 +1,7 @@
+if RUBY_VERSION < '1.9.0' # this is a string comparison, but it should work for any old ruby version
+  $stderr.puts "Heroku Toolbelt requires Ruby 1.9+."
```
wuputah: Would it make sense to `exit(1)` as well?

dickeyxxx: added

coryschires:
+1 

Any idea when this will be merged? I'm unable to upgrade to the latest version due to #1474

dickeyxxx:
it's looking like Monday. I need to reach out to some people that are still using 1.8.7

dickeyxxx:
I added https://github.com/heroku/heroku/pull/1482 to warn people now

PR #1482, opened 2015-04-02
added warning for users on ruby 1.8.7

dickeyxxx:


lib/heroku/cli.rb
```diff
@@ -1,3 +1,8 @@
+if RUBY_VERSION < '1.9.0'
+  $stderr.puts "WARNING: Heroku Toolbelt will require Ruby 1.9+ beginning next week"
+  $stderr.puts "https://github.com/heroku/heroku/pull/1479"
```
raulb: @dickeyxxx I would have added the actual date (in case someone see this message after the next week). Or, maybe accommodate the message to be more flexible. Something like:

> WARNING: Heroku Toolbelt will start requiring Ruby 1.9+ on MM/DD/YYYY.

Unless, next week you'll update this message to something like:

> WARNING: Heroku Toolbelt requires Ruby 1.9+.

dickeyxxx: :+1: @craigkerstiens mentioned the same thing. I already pushed this out and sent an email to all the users I found that used the CLI in the last week under 1.8.7. I'm going to see how it plays out a bit. (Leaving the date out was semi-intentional to give myself some room, but in hindsight I should've just set a solid date)

dickeyxxx: The date right now is Wednesday

raulb: Awesome. And for the record, I think it was the right call to leave the date out. I was mostly wondering how it'd look like after the next week.

dickeyxxx: oh! ha. I didn't think of that.

dickeyxxx: I updated this in #1484

raulb: @dickeyxxx :+1:

PR #1485, opened 2015-04-02
require the at option when scheduling backups

neovintage:
Small UX problem whereby someone creates a schedule and then issues the same command without the at flag thinking they could view schedules.  The schedule is overwritten and set back to 4:00 UTC.

/cc @uhoh-itsmaciek @chadbailey59 

What do you guys think?

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -396,7 +396,10 @@ EOF
   def schedule_backups
     db = shift_argument
     validate_arguments!
-    at = options[:at] || '04:00 UTC'
+    if !options[:at]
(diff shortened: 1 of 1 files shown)


uhoh-itsmaciek:
Interesting idea. It seems weird to have required flags (options are usually... optional), but it might make sense here. Are there other instances of this behavior in the CLI?

dickeyxxx:
perhaps prompt if no time is scheduled instead of forcing them to run the command again?

neovintage:
@uhoh-itsmaciek I quickly grep-ed across codebase and didn't see it being done anywhere else.  I wouldn't want to be the first one to start requiring options.

@dickeyxxx That's a good idea.

PR #1495, opened 2015-04-09
Add `-s` to `pg:backups public-url` so people feel better about scripting

chadbailey59:
This adds an option to force hiding the "This link will expire in..." text from `pg:backups public-url`. This isn't entirely necessary, as we already check for TTY, but [ruby can be tricky with that sometimes anyway](https://gist.github.com/timuruski/6072363). I'll be able to shut up more than a few support tickets with this.

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -39,6 +39,7 @@ class Heroku::Command::Pg < Heroku::Command::Base
   #  capture DATABASE               # capture a new backup
   #  restore [[BACKUP_ID] DATABASE] # restore a backup (default latest) to a database (default DATABASE_URL)
   #  public-url BACKUP_ID           # get secret but publicly accessible URL for BACKUP_ID to download it
+  #    -s, --suppress               #   Suppress expiration message (for use in scripts)
   #  cancel                         # cancel an in-progress backup
   #  delete BACKUP_ID               # delete an existing backup
   #  schedule DATABASE              # schedule nightly backups for given database
@@ -431,7 +432,7 @@ EOF
     end
 
     url_info = client.transfers_public_url(backup_num)
-    if $stdout.tty?
+    if $stdout.tty? && !options[:suppress]
       display <<-EOF
 The following URL will expire at #{url_info[:expires_at]}:
   "#{url_info[:url]}"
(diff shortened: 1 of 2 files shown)


uhoh-itsmaciek:
If we add this, we should probably call this `-q` / `--quiet` for consistency with UNIX tools (to the extent that there is any).

will:
Could also move the text to stderr

uhoh-itsmaciek:
Incidentally, the linked TTY stuff is pretty straightforward (if not intuitive): a child process inherits stdin, stdout, and stderr from its parent. So it doesn't really matter where you're calling a process--only how you're redirecting output. So if you're going to be using the output somewhere that's not a TTY (e.g., redirecting it to another command, storing it in a variable, or using the output of a subshell invocation directly), it should just work correctly.

chadbailey59:
@uhoh-itsmaciek , I changed it to quiet. (I didn't like suppress anyway.)

@will re: our discussion in HipChat: I agree this isn't necessary because of the TTY stuff. But we've had `public-url` in the wild for a day now, and I already got a ticket about how "this extra line of text is going to break my scripts", and I know it won't be the first. Honestly, it was quicker for me to write this code than to figure out how to fit an explanation of the TTY behavior into `heroku help pg:backups public-url`.

That said, if consensus is that this is a bad idea, I'm happy to trash it.

uhoh-itsmaciek:
@chadbailey59 there are far worse command flags out there =)

I guess people may have incorrectly assumed that the extra quotes in the old command's output wouldn't break anything (when in fact things were not-breaking because of the special `#tty?` handling).

If it reduces support burden, I'm all for it. One could conceivably want terse output even if output *is* a TTY (say, to more easily copy and paste into something).

lib/heroku/command/pg_backups.rb
```diff
@@ -39,6 +39,7 @@ def copy
   #  capture DATABASE               # capture a new backup
   #  restore [[BACKUP_ID] DATABASE] # restore a backup (default latest) to a database (default DATABASE_URL)
   #  public-url BACKUP_ID           # get secret but publicly accessible URL for BACKUP_ID to download it
+  #    -q, --quiet                  #   Hide expiration message (for use in scripts)
```
dickeyxxx: this doesn't appear to be in the right place

dickeyxxx: oops, nevermind, it is

PR #1501, opened 2015-04-10
workaround for ruby bug of home directory on windows

dickeyxxx:


joaoportela:
This workaround isn't working for me. My home dir is `c:/Users/João`.

If we look at what `Dir.home` returns, like so:

    c:\Program Files (x86)\Heroku\ruby-1.9.3\bin>irb.bat
    irb(main):001:0> Dir.home
    => "C:/Users/Jo\xC6o"

We can see that the `ã` is encoded as `C6`, by looking at the (charset info)[http://www.fileformat.info/info/unicode/char/e3/charset_support.htm] we see that on my machine it must be one of: IBM00858, IBM850, IBM857. I tried IBM857 and it worked.

I also encoded it as utf-8 so that `formatted_error.join("\n")` doesn't fail if it is called. I ended up with this line:

    Dir.home.force_encoding('ibm857').encode("utf-8")

Thanks.

dickeyxxx:
apologies @joaoportela this bug seems to keep returning :rage: 

I think this might fix it https://github.com/heroku/heroku/pull/1744

can you tell me what `File.expand_path('~')` returns on your machine?

joaoportela:
Here it is:

    irb(main):006:0> RUBY_VERSION
    => "1.9.3"
    irb(main):007:0> File.expand_path('~')
    => "C:/Users/Jo\xE3o"
    irb(main):008:0> File.expand_path('~').encoding
    => #<Encoding:Windows-1252>

#1744 does seem to have fixed it.

dickeyxxx:
great!

PR #1504, opened 2015-04-14
add method that should have been added with cafe9ada0df64abec36d59da549b...

will:
...ac952263beb9 back in July 2014

### DIFF
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -495,6 +495,10 @@ private
     display
   end
 
+  def in_maintenance?(app)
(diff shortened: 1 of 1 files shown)


will:
cc @uhoh-itsmaciek @heroku/department-of-data @chadbailey59

dickeyxxx:
wow that took a long time to catch!

will:
Yeah, the method is defined in pg-extras, and there is a high intersection with people who manage their own maintenance windows, and have that plugin installed. Took this long to get a customer who didn't have both.

PR #1513, opened 2015-04-24
Additional output for `pg:backups` commands

chadbailey59:
This PR ensures we're outputting as much info as we can for `pg:backups` and `pg:backups info` commands.


chadbailey59:
Also, two bug fixes:

1. Restores that came from `htcat` wouldn't show in restores.
2. If an app had no backups but did have restores/copies, they wouldn't show because 'no backups' was using the `error` helper, which (helpfully) exits.

neovintage:
@chadbailey59 I tested out the code.  I have a couple of comments. 

* when you do pg:copy of databases across applications,  there's no way of knowing that it happened.  You just see two different colors.  
* The output for copies can be exceptionally long.  Is there a way to limit that to maybe the last 10?

I have screenshots if you need them.

lib/heroku/command/pg_backups.rb
```diff
@@ -399,7 +422,7 @@ def poll_transfer(action, transfer_id)
       redisplay <<-EOF
 An error occurred and your backup did not finish.
 
-Please run `heroku logs --ps pg-backups` for details.
+Please run `heroku pg:backups info #{transfer_name(backup)}` for details.
```
uhoh-itsmaciek: Yes thank you. I've been meaning to do this for forever and keep forgetting.

.ruby-version
```diff
@@ -0,0 +1 @@
+2.2.0
```
uhoh-itsmaciek: This should probably be separate or just skipped.

uhoh-itsmaciek:
Looks great. I haven't tried it, but I think your changes look pretty straightforward, so with you and Rimas giving it a shot (and the specs) I'm sure it's fine. Let me know if you want me to run through something in particular.

I'm neutral on the log output @neovintage brought up--that can certainly be a lot of output, but you're generally only looking when you want to know so maybe that's fine? I don't follow the cross-app copy issue. Maybe the screenshot would help.

neovintage:
The setup is I have two apps, neovintage and pgbackups-rims.  I copied gold
from neovintage to Blue on pgbackups-rims.  When I look at this output on
pgbackups-rims, I only see GOLD to BLUE.  I don't have context that it came
from another app.

neovintage:
Screenshot
screen shot 2015-04-27 at 9 14 45 pm

uhoh-itsmaciek:
Got it. I think I know how to fix that, but maybe we should address that independently of these changes so these can go out more quickly...

chadbailey59:
I removed the ruby-version file. I also limited the restore and copy outputs to the 20 most recent.

I absolutely think @neovintage 's concern about cross-app restores is valid. I noticed it too when I was testing, but I'm trying to keep this as limited in scope as possible to get the bugfixes out.

neovintage:
Is that 20 total across both or each?  I would suggest limiting to last
10 in each.

What about the web interface to all of this?

chadbailey59:
I just pushed a commit limiting to the last 10 operations each. It'll show
10 most recent restores and 10 most recent pg:copy operations.

As far as the web, I'm not sure.

neovintage:
Thanks for the fix.  We can push this out now.  We'll have to schedule some time to get PHC fixed to include cross app copies and restores.  For future reference, we need to make sure that if we make changes to CLI we evaluate if a change is necessary to PHC.  

/cc @chadbailey59 @uhoh-itsmaciek

PR #1519, opened 2015-04-29
added --exit-code flag for heroku run

dickeyxxx:


dickeyxxx:
need to handle stdin and terminal sizes (at least) before I can merge this

PR #1528, opened 2015-05-05
move to v4 run implementation

dickeyxxx:


whatthewhat:
@dickeyxxx I think this change breaks `heroku-repo` (https://github.com/heroku/heroku-repo/issues/14#issuecomment-143844171 as an example).

Is there a migration strategy for this I can read somewhere? I'd be happy to submit a PR to `heroku-repo`.

PR #1531, opened 2015-05-05
Retry backup status polling for a while

uhoh-itsmaciek:
We sometimes have intermittent issues with API endpoint failures
for checking backup status; we're working to fix these, but in
the meantime, there's no reason to fail a command polling transfer
status when this happens; just ignore the update and try again
at the next polling interval.

/cc @chadbailey59


dickeyxxx:
lgtm

PR #1534, opened 2015-05-06
v4 version of heroku status

dickeyxxx:
cc @raulb

we'll need to change it from _status to status on the npm package before merging this


raulb:
@dickeyxxx Published!

PR #1539, opened 2015-05-07
Fix unscheduling backups for some config var configurations

uhoh-itsmaciek:
For some applications with aliased config vars, it's possible
that the unschedule command can be too stringent in verifying
the database to unschedule, making it impossible to unschedule
backups without manually calling the API.

This makes the unschedule verification much simpler.

/cc @chadbailey59 @neovintage @hgmnz

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -44,7 +44,7 @@ class Heroku::Command::Pg < Heroku::Command::Base
   #  delete BACKUP_ID               # delete an existing backup
   #  schedule DATABASE              # schedule nightly backups for given database
   #    --at '<hour>:00 <timezone>'  #   at a specific (24h clock) hour in the given timezone
-  #  unschedule DATABASE            # stop nightly backup for database
+  #  unschedule SCHEDULE            # stop nightly backup for database
   #  schedules                      # list backup schedule
   def backups
     if args.count == 0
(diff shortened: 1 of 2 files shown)


hgmnz:
woot woot. Thanks @uhoh-itsmaciek! 

Yeah, much simpler and also you know, correct. So +1!

lib/heroku/command/pg_backups.rb
```diff
@@ -513,12 +513,15 @@ def unschedule_backups
     db = shift_argument
     validate_arguments!
 
+    if db.nil?
+      abort("Must specify database to unschedule backups for")
+    end
```
hgmnz: If there is only one schedule, should we accept no `db` here?

For this message, I'd add a hint: "Must specify a database. Run `heroku help pg:backups` for usage information."

uhoh-itsmaciek: I think it does make sense to require an argument. Maybe instead of pointing people to `help`, we should just list what's available to unschedule instead?

chadbailey59: I agree we should require an argument. :)

uhoh-itsmaciek:
@hgmnz @chadbailey59 Improved the error message / error handling. How's this look?

tef:
Just got a ticket for this—any sign of merging?

neovintage:
@uhoh-itsmaciek @tef  The error messages and handling look good to me.

PR #1554, opened 2015-05-08
Bug fixes for users with partial privileges on apps:info

kahlouie:
Waiting on @balansubr's okay before it's ready to merge.

@dickeyxxx are there any tests I should add?


dickeyxxx:
nope, travis will take care of that

balansubr:
Looks good to me for now. If we get more customer feedback that some of these should not be displayed for users with view privilege, we will adjust then.

kahlouie:
@dickeyxxx ready to merge

PR #1557, opened 2015-05-12
pg:promote: work with config var names again and create JIT backup attachments for outgoing database

bjeanes:
This PR has 2 distinct changes:

* Fix regression preventing promotion with full config var name (e.g. `heroku pg:promote HEROKU_POSTGRESQL_CYAN_URL`)
* Avoid "can't remove last attachment" error message when promoting by ensuring there is at least 1 alternate attachment for the add-on which currently occupies the `DATABASE` alias.

E.g., given: 

```sh-session
$ heroku addons:create heroku-postgresql -a mighty-mesa-1351
Creating helping-busily-4207... done
Adding helping-busily-4207 to mighty-mesa-1351... done
Setting DATABASE_URL and restarting mighty-mesa-1351... done, v3
Database has been created and is available
 ! This database is empty. If upgrading, you can transfer
 ! data from another database with pgbackups:restore
Use `heroku addons:docs heroku-postgresql` to view documentation.

$ heroku addons:create heroku-postgresql -a mighty-mesa-1351
Creating cooling-keenly-1251... done
Adding cooling-keenly-1251 to mighty-mesa-1351... done
Setting HEROKU_POSTGRESQL_BRONZE_URL and restarting mighty-mesa-1351... done, v4
Database has been created and is available
 ! This database is empty. If upgrading, you can transfer
 ! data from another database with pgbackups:restore
Use `heroku addons:docs heroku-postgresql` to view documentation.
```

Before second change:

```sh-session
$ heroku pg:promote BRONZE -a mighty-mesa-1351
Promoting cooling-keenly-1251 to DATABASE_URL on mighty-mesa-1351... failed
 !    DATABASE on mighty-mesa-1351 is the last billing attachment to helping-busily-4207; either remove the add-on entirely or add a new attachment to it before replacing its config vars.
```

After second change:

```sh-session
$ heroku pg:promote HEROKU_POSTGRESQL_IVORY_URL -a mighty-mesa-1351
Ensuring an alternate alias for existing DATABASE... done, HEROKU_POSTGRESQL_AQUA
Promoting reading-fleetingly-3902 to DATABASE_URL on mighty-mesa-1351... done
```

The resource behind `DATABASE` had no other attachments, so it created `HEROKU_POSTGRESQL_AQUA` so that it could be replaced. I think this change is good, however it does have one rather large caveat which could affect users of the `preboot` labs feature in particular: it creates 2 releases (one to add an attachment, the other to replace an attachment) which could trigger 2 subsequent restarts. ccing @heroku/department-of-data for comment here.

We could alternatively pursue improvements to the error message that indicate the action that the user can take for correction. However, such an action would also cause 2 restarts.

### DIFF
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -85,6 +85,7 @@ class Heroku::Command::Pg < Heroku::Command::Base
     end
     validate_arguments!
 
+    db = db.sub(/_URL$/, '') # allow promoting with a var name
     addon = resolve_addon!(db)
 
     attachment_name = 'DATABASE'
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -86,7 +86,7 @@ class Heroku::Command::Pg < Heroku::Command::Base
     validate_arguments!
 
     db = db.sub(/_URL$/, '') # allow promoting with a var name
-    addon = resolve_addon!(db)
+    addon = resolve_addon!(db) { |addon| addon['addon_service']['name'] == 'heroku-postgresql' }
 
     promoted_name = 'DATABASE'
 
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -88,14 +88,26 @@ class Heroku::Command::Pg < Heroku::Command::Base
     db = db.sub(/_URL$/, '') # allow promoting with a var name
     addon = resolve_addon!(db)
 
-    attachment_name = 'DATABASE'
-    action "Promoting #{addon['name']} to #{attachment_name}_URL on #{app}" do
+    promoted_name = 'DATABASE'
+
+    action "Ensuring an alternate alias for existing #{promoted_name}" do
+      backup = find_or_create_non_database_attachment(app)
+
+      if backup
+        @status = backup['name']
+      else
+        @status = "not needed"
+      end
+
+    end
+
+    action "Promoting #{addon['name']} to #{promoted_name}_URL on #{app}" do
       request(
         :body     => json_encode({
           "app"     => {"name" => app},
           "addon"   => {"name" => addon['name']},
           "confirm" => app,
-          "name"    => attachment_name
+          "name"    => promoted_name
         }),
         :expects  => 201,
         :method   => :post,
@@ -667,5 +679,41 @@ private
     # some people alais psql, so we need to find the real psql
     # but windows doesn't have the command command
     running_on_windows? ? 'psql' : 'command psql'
+  end
+
+  # Finds or creates a non-DATABASE attachment for the DB currently
+  # attached as DATABASE.
+  #
+  # If current DATABASE is attached by other names, return one of them.
+  # If current DATABASE is only attachment, create a new one and return it.
+  # If no current DATABASE, return nil.
+  def find_or_create_non_database_attachment(app)
+    attachments = get_attachments(:app => app)
+
+    current_attachment = attachments.detect { |att| att['name'] == 'DATABASE' }
+    current_addon      = current_attachment && current_attachment['addon']
+
+    if current_addon
+      existing = attachments.
+        select { |att| att['addon']['id'] == current_addon['id'] }.
+        detect { |att| att['name'] != 'DATABASE' }
+
+      return existing if existing
+
+      # The current add-on occupying the DATABASE attachment has no
+      # other attachments. In order to promote this database without
+      # error, we can create a secondary attachment, just-in-time.
+      request(
+        # Note: no attachment name provided; let the API choose one
+        :body     => json_encode({
+          "app"     => {"name" => app},
+          "addon"   => {"name" => current_addon['name']},
+          "confirm" => app
+        }),
+        :expects  => 201,
+        :method   => :post,
+        :path     => "/addon-attachments"
+      )
+    end
   end
 end
--- a/lib/heroku/helpers/addons/resolve.rb
+++ b/lib/heroku/helpers/addons/resolve.rb
@@ -16,8 +16,8 @@ module Heroku::Helpers
       # Finds attachments that match provided identifier.
       #
       # Always returns an Array of 0 or more results.
-      def resolve_attachment(identifier)
-        case identifier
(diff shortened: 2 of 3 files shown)


dickeyxxx:
looks good from my end, test failure though

neovintage:
@bjeanes In the Redis code, a user can instantiate the first redis instance and that get's set to `REDIS_URL`, any subsequent redis instance that's created gets a `COLOR_URL`.  We have a promote command in there as well, the instance on `REDIS_URL` then gets a `COLOR_URL` and the instance getting promoted gets set to REDIS_URL.  That gets rid of the "can't remove last attachment" error.  Is that code causing 2 restarts as well?

bjeanes:
@neovintage you'll need to point me to the implementation, but I'm almost certain it would be. This is one of the motivators behind my server-side promote RFC: https://github.com/heroku/rfcs/pull/7, which you've seen.

bjeanes:
@dickeyxxx @neovintage the behaviour you are describing is actually identitical to what I'm doing, except that I am not giving the current one a color URL if it already has another attachment. Ideally, we could factor these two commands into a shared implementation. Other than the attachment name, they should be identical, I think...

bjeanes:
@dickeyxxx yup saw the test failure. If this change gets a ~+1 from @heroku/department-of-data I'll fix that up without hesitation.

lib/heroku/command/pg.rb
```diff
@@ -85,16 +85,29 @@ def promote
...
     validate_arguments!
 
+    db = db.sub(/_URL$/, '') # allow promoting with a var name
     addon = resolve_addon!(db)
 
-    attachment_name = 'DATABASE'
-    action "Promoting #{addon['name']} to #{attachment_name}_URL on #{app}" do
+    promoted_name = 'DATABASE'
+
+    action "Ensuring an alternate alias for existing #{promoted_name}" do
```
uhoh-itsmaciek: This is kind of misleading and potentially confusing in the (common?) case when there is no existing attachment, no?

lib/heroku/command/pg.rb
```diff
@@ -667,4 +680,40 @@ def psql_cmd
...
+  # If no current DATABASE, return nil.
+  def find_or_create_non_database_attachment(app)
+    attachments = get_attachments(:app => app)
+
+    current_attachment = attachments.detect { |att| att['name'] == 'DATABASE' }
+    current_addon      = current_attachment && current_attachment['addon']
+
+    if current_addon
+      existing = attachments.
+        select { |att| att['addon']['id'] == current_addon['id'] }.
```
uhoh-itsmaciek: There's always an `addon`, right? We don't have to worry about `nil` here?

lib/heroku/command/pg.rb
```diff
@@ -85,16 +85,29 @@ def promote
...
-    action "Promoting #{addon['name']} to #{attachment_name}_URL on #{app}" do
+    promoted_name = 'DATABASE'
+
+    action "Ensuring an alternate alias for existing #{promoted_name}" do
+      backup = find_or_create_non_database_attachment(app)
+
+      if backup
+        @status = backup['name']
+      else
+        @status = "not needed"
```
uhoh-itsmaciek: Does `@status` tie into `action`? How is this communicated to the user?

lib/heroku/command/pg.rb
bjeanes: Mayeb, but the alternative is the an error. We could work on making the error message instead.

lib/heroku/command/pg.rb
bjeanes: You mean `att['addon']`? That's never nil.

lib/heroku/command/pg.rb
bjeanes: It suffixes the `...done` message for the action. It used to replace it, but now it'd look like:

```
Ensuring an alternate alias for existing DATABASE... done, HEROKU_POSTGRESQL_AQUA
```

uhoh-itsmaciek: Oh okay. I think in that case the message is actually reasonably clear...

neovintage:
@bjeanes 

I was going through some edge cases on this and I ran into this situation:
http://d.pr/i/1hUFH

I had a JADE color on a redis instance and postgres when this error was thrown.  Is is possible to scope the check just against the postgres attachments?

dickeyxxx:
on an unrelated note, I don't really like the hyphens on that list. I think it would be just as readable without. I might change that a bit when we move this stuff to v4. Maybe change it out for cyan colored labels

bjeanes:
@neovintage Nice catch. Yeah that should be possible. I'll look into it!

bjeanes:
@neovintage fixed with 9eddfcf and 50575a6

Taytay:
:+1: for this PR. Thanks!
(Our builds are broken until this gets merged, so we are watching it anxiously.)

bjeanes:
@Taytay because of the `_URL`?

Taytay:
@bjeanes This is basically going to fix the error we are running into #1562 
```
00:26:23.896 + heroku pg:promote YELLOW --app our-new-app
00:26:24.809 Promoting thinking-gently-8360 to DATABASE_URL on our-new-app... failed
00:26:24.810  !    DATABASE on our-new-app is the last billing attachment to sailing-keenly-4016; either remove the add-on entirely or add a new attachment to it before replacing its config vars.
```
Our next step in our build (after this promotion) is to delete that old database, but it looks like we'll need to add an attachment right before we promote in order for this to work, at least until this PR lands. Is that right?

bjeanes:
@Taytay ah right. Yes. You could manually create the other attachment for now to get your builds going again, if needed.

Taytay:
Cool - am doing so now. Thanks.

bjeanes:
@dickeyxxx I checked the code for the `heroku-redis` promote and it also has the double attach issue. IMO, this should be OK for merge. If we want to find an alternate way to avoid the restarts, it can be done in a way that addresses both. If you want to wait for an explicit +1 from DoD, then I suggest we at least merge in the first commit now but otherwise I'm comfortable with this going in for a merge.

PR #1559, opened 2015-05-12
ensure that polling has a value for backup in exception tracker

dickeyxxx:
https://rollbar.com/heroku-cli/heroku-ruby-cli/items/13733/occurrences/5262074896/


dickeyxxx:
cc @uhoh-itsmaciek

uhoh-itsmaciek:
Good call. I think this could be safer still, but this guards against a common error; I'll follow up when I have time.

PR #1588, opened 2015-05-26
Tell you where youre copying to and from in pg:copy

neovintage:
Need to tell devs where you're copying to and from.  It's a little scary to issue the command without some kind of idea of what's going to happen.

@chadbailey59 does the copy look good?


neovintage:
@dickeyxxx - Updated language in output to be more clear about what's going on.

PR #1589, opened 2015-05-27
Domain CNAME Targets

ryanbrainard:
This change adds the following:
 - Splits the domains output into two tables for the standard `Development Domain` and `Custom Domains` (matching pending changes in Dashboard)
 - Adds a new `CNAME Target` column for custom domains
 - Adds instructions to use CNAME target when adding a custom domain
 - Uses `v3.domain-cname` variant for the commands above

cc: @dickeyxxx @tim-lang @itchymutt

### DIFF
--- a/lib/heroku/command/domains.rb
+++ b/lib/heroku/command/domains.rb
@@ -48,14 +48,19 @@ module Heroku::Command
     # $ heroku domains:add example.com
     # Adding example.com to example... done
     #
+    # !   Configure your application’s DNS to point to example.herokudns.com
+    # !   For help with custom domains, see https://devcenter.heroku.com/articles/custom-domains
+    #
     def add
       unless domain = shift_argument
         error("Usage: heroku domains:add DOMAIN\nMust specify DOMAIN to add.")
       end
       validate_arguments!
-      action("Adding #{domain} to #{app}") do
-        api.post_domain(app, domain)
+      domain = action("Adding #{domain} to #{app}") do
+        api.post_domains_v3_domain_cname(app, domain).body
       end
+      output_with_bang "Configure your application's DNS to point to #{domain['cname']}"
+      output_with_bang "For help, see https://devcenter.heroku.com/articles/custom-domains"
     end
 
     # domains:remove DOMAIN
--- a/lib/heroku/command/domains.rb
+++ b/lib/heroku/command/domains.rb
@@ -18,7 +18,7 @@ module Heroku::Command
     # example.herokuapp.com
     #
     # === Custom Domains
-    # Domain Name  CNAME Target
+    # Domain Name  DNS Target
     # -----------  ---------------------
     # example.com  example.herokudns.com
     #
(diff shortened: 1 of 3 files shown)


ryanbrainard:
I did some digging and it is there is small handful of apps that do not have a development domain. I added back the error handling for that case and tweaked the output when no custom domains are found to show the user how to add one, which is similar to how `heroku certs` behaves and looks better with the two table design.

ryanbrainard:
@dickeyxxx The API changes have been merged and deployed. This should be good to go from my end now.

lib/heroku/command/domains.rb
```diff
@@ -13,17 +14,35 @@ class Domains < Base
     #Examples:
     #
     # $ heroku domains
-    # === Domain names for example
-    # example.com
+    # === Development Domain
+    # example.herokuapp.com
+    #
+    # === Custom Domains
+    # Domain Name  CNAME Target
```
bjeanes: `CNAME`/`ALIAS` ?

ryanbrainard: there was a lot of discussion about the wording for this column in Dashboard, and this was copied from there. @tim-lang, @itchymutt: thoughts? I don't think we should block on this, but worth consideration for a follow up.

tim-lang: For the UI, we were looking for a name that implied it was the "right hand
side" of the CNAME record, e.g., the canonical name, not one of several
nicknames/aliases/etc.

Don't think we should block either but let me know if there are strong
opinions or I am misunderstanding. :)

tim-lang: "Domain Name" and "CNAME target" look good to me as column headers.
"Development Domain" needs to be considered further.

bjeanes: But `CNAME` implies using that DNS record type, which you shouldn't in all cases. `Record Target` or something more generic seems more sensible to me.

ryanbrainard: I think 'DNS Record Target' would capture everyone's concerns. It works for both CNAME and ALIAS, speaks to the right hand side of the record, and hints that it is for DNS. If we change it here, we also need to update Dashboard.

sclasen: +1 that is the phrase I had in mind too

ryanbrainard: @jesperfj, @itchymutt: are you ok with using `DNS Record Target` both here and Dashboard?

jesperfj: What's wrong with CNAME target?

ryanbrainard: @jesperfj see the thread above

jesperfj: I see. Sorry was on email. It's to not exclude ALIAS records. Fair enough.
Do we need "Record"? how about just "DNS Target"?

ryanbrainard: Sounds good. Making the change here. This :bike:-shed is closed.

ryanbrainard:
We just had an offline discussion of these labels. Closing this pull request for now while we fix things up. Will re-open when it's ready to go.

PR #1593, opened 2015-05-29
Use created_at to find last backup, not number

tef:
We use the backup number, which doesn't work when there are imported older backups, and they propagate to the top.


uhoh-itsmaciek:
Looks fine, but is there a problem with the current mechanism?

tef:
Yep, if you sort by number, the letter prefixes ruin it. I have an open ticket linked in trello for this

uhoh-itsmaciek:
Ah, got it. +1.

PR #1595, opened 2015-05-29
Allow client to cancel by transfer name

tef:
Allow pg:backups cancel to take an optional transfer name, and change the default for when no name provided to pick the newest active backup, rather than relying on order from api.

@uhoh-itsmaciek this seem ok?


uhoh-itsmaciek:
Sure, looks good.

Issue #1586, opened 2015-05-26
ps:scale does not understand the new dyno types


shedd:
There seems to be a bug in trying to scale applications using the `ps:scale` command with the new dyno types.

I keep hitting this error:

```
$ heroku ps:scale web=1:Standard-2X worker=1:Standard-1X --app appname
! Usage: heroku ps:scale DYNO1=AMOUNT1[:SIZE] [DYNO2=AMOUNT2 ...]
! Must specify DYNO and AMOUNT to scale.
$ heroku ps:scale web=1:standard-2X worker=1:standard-1X --app appname
! Usage: heroku ps:scale DYNO1=AMOUNT1[:SIZE] [DYNO2=AMOUNT2 ...]
! Must specify DYNO and AMOUNT to scale.
```

This syntax works with the old 1X, 2X, PX dynos, but not the new names.

bharat:
In some cases I can use `ps:type` to change the type, eg:

```
$ heroku ps:type --app foo-bar-app Free
Changing dyno type... done.
dyno  type  qty  cost/mo
----  ----  ---  -------
web   Free    1        0
```

But in some cases I can't:

```
$ heroku ps:type --app other-similar-app Free
Changing dyno type... failed
 !    Can not set dyno type on this app. Please use `heroku ps:scale` to change process size and scale.
```

What's the right approach?

bharat:
More info in my case - once I use the web interface to change the dyno type (even if I change it back to the original value) then heroku toolbelt works on it. So if I go to `other-similar-app` from the example above and set the dyno type to Hobby, save, then set it back to Free and save and *then* run the command again it works.

bharat:
I deleted my earlier comments - they were about an unrelated issue (you can't use heroku toolbelt to change the dyno type until you take an action in the web ui first).

However - I believe this issue is caused by a failure in the regex here:
https://github.com/heroku/heroku/blob/master/lib/heroku/command/ps.rb#L232

My guess is that the `\w+` needs to be `[\w-]+`.

dickeyxxx:
fixed in 3.37.4

PR #1596, opened 2015-05-30
Improve 'ps scale' to handle dyno types that have non-alphanumeric

bharat:
characters in them (eg: Standard-1X)

Fix for #1586.


bharat:
Sorry for the lack of tests - I haven't written Ruby tests in a long while. If you point me at some docs to run tests and submit pull requests with good hygiene I'll be happy to do it the right way.

dickeyxxx:
thanks!

Issue #1149, opened 2014-07-22
auth:whoami should fail if not logged in


listrophy:
There doesn't seem to be any way in the entire toolbelt to determine _whether_ I'm logged in without triggering an interactive login.

I'd like for `heroku auth:whoami` to return with exit status > 0 when not logged in. The actual output in such a case is up to you.

Context: I'm writing a script that calls `heroku config`, and I'd like to provide a helpful error message.

camstuart:
Agreed. I'm trying to automate the login process, but if there is no current user logged in, an ```auth:login``` triggers. Which is a bit annoying

dickeyxxx:
https://github.com/heroku/heroku/pull/1598

PR #1598, opened 2015-06-01
use v4 for auth:whoami

dickeyxxx:


mikehale:
I think this is doing the wrong thing. We should not run `whoami` we should lookup the logged in user based on what is in `~/.netrc`.

dickeyxxx:
why?

mikehale:
That's what it used to do no?

dickeyxxx:
it did, but that seems wrong to me. The login field in netrc could be anything, it wouldn't necessarily match to the `HEROKU_API_KEY` and it might have an expired token. This ticket mentions some of the issues with the old functionality: https://github.com/heroku/heroku/issues/1149

mikehale:
If I wanted to see my local machine user I'd just do `echo $USER` or `whoami`. If I want to see who I am on heroku I run `heroku auth:whoami`.

... just tried this locally and I'm definitely getting back my heroku user. I guess I was thinking that these were equivalent, but it seems to be doing something else which I don't understand:

```
Heroku::JSPlugin.run('whoami', nil, ARGV[1..-1])
```

and 

```
whoami
```

dickeyxxx:
oh so when I do `Heroku::JSPlugin.run('whoami', nil, ARGV[1..-1])` what that does is call `heroku whoami` on the v4 cli at `~/.heroku/heroku-cli`. Note that I changed this after the PR to be `~/.heroku/heroku-cli auth:whoami`

dickeyxxx:
but yeah, the new functionality calls `/account` on the api and returns the email there. This way should be better for scripting though because it's not just reading the .netrc file, it's actually checking to see if you will be able to access the API

dickeyxxx:
some users were surprised that even though `heroku auth:whoami` returned an email that didn't mean they could actually use the API

mikehale:
Ah, that makes sense.

PR #1603, opened 2015-06-03
Sleep longer to avoid exhausting api quota

tef:
/cc @uhoh-itsmaciek

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -426,7 +426,7 @@ EOF
           raise
         end
       end
-      sleep 1
+      sleep 3
     end until backup[:finished_at]
     if backup[:succeeded]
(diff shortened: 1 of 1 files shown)


uhoh-itsmaciek:
+1

For context, calls to `postgres-api.heroku.com` must make a call as the calling user to authenticate (because addons don't really have a better auth mechanism). For users polling long-running backups, this can have a significant impact on their API usage vis-à-vis rate limits. We probably need to come up with a better authentication mechanism in general, but this should be a reasonable triage.

bjeanes:
> We probably need to come up with a better authentication mechanism in general

:+1:. If you have any ideas, I'd love to chat about this in the next couple of weeks when I'm in town. Next quarter we're likely working on a new integration and I'd love to incorporate auth changes into it if it makes sense.

PR #1604, opened 2015-06-04
Use correct app name in warnings, allow using app::backup for restore source

tef:
/cc @uhoh-itsmaciek

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -30,7 +30,7 @@ class Heroku::Command::Pg < Heroku::Command::Base
       affected = target.name.downcase
     else
       confirm_with = target.attachment.app
-      affected = "the app #{target.attachment.app}"
+      affected = "the app: #{target.attachment.app}"
     end
 
     message = "WARNING: Destructive Action"
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -23,12 +23,22 @@ class Heroku::Command::Pg < Heroku::Command::Base
 
     attachment = target.attachment || source.attachment
 
+    if target.attachment.nil?
+      target_url = URI.parse(target.url)
+      confirm_with = target_url.path[1..-1]
+      confirm_with = target_url.host if confirm_with.empty?
+      affected = target.name.downcase
+    else
+      confirm_with = target.attachment.app
+      affected = "the app #{target.attachment.app}"
+    end
+
     message = "WARNING: Destructive Action"
     message << "\nThis command will remove all data from #{target.name}"
     message << "\nData from #{source.name} will then be transferred to #{target.name}"
-    message << "\nThis command will affect the app: #{attachment.app}"
+    message << "\nThis command will affect #{affected}"
 
-    if confirm_command(attachment.app, message)
+    if confirm_command(confirm_with, message)
       xfer = hpg_client(attachment).pg_copy(source.name, source.url,
                                             target.name, target.url)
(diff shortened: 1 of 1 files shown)


lib/heroku/command/pg_backups.rb
```diff
@@ -26,9 +26,9 @@ def copy
     message = "WARNING: Destructive Action"
     message << "\nThis command will remove all data from #{target.name}"
     message << "\nData from #{source.name} will then be transferred to #{target.name}"
-    message << "\nThis command will affect the app: #{app}"
+    message << "\nThis command will affect the app: #{attachment.app}"
 
-    if confirm_command(app, message)
+    if confirm_command(attachment.app, message)
```
tef: I think I might need to fix the logic here.

lib/heroku/command/pg_backups.rb
```diff
@@ -380,16 +384,16 @@ def restore_backup
...
-        abort("No backups. Capture one with `heroku pg:backups capture`.")
+        abort("No backups for #{backup_app}. Capture one with `heroku pg:backups capture`.")
       elsif backup.nil?
-        abort("Backup #{backup_name} not found.")
+        abort("Backup #{backup_name} not found for #{backup_app}.")
       elsif !backup[:succeeded]
-        abort("Backup #{backup_name} did not complete successfully; cannot restore it.")
+        abort("Backup #{backup_name} for #{backup_app} did not complete successfully; cannot restore it.")
       end
       restore_url = backup[:to_url]
```
tef: This url should be fine if we retrieve it from another app? Or will we need to resolve it to a public url ?

uhoh-itsmaciek: Let's just resolve it so we can tighten the group restrictions.

lib/heroku/command/pg_backups.rb
uhoh-itsmaciek: Not `target.attachment.app`?

uhoh-itsmaciek: And ugh, great catch.

tef: the target could have no attachment? would we want a warning then ?

tef:
@uhoh-itsmaciek 

Fixes:

- EU Timezones, Zulu Timezone
- Bad warnings, should always show target database if specified in copy, or restore
- Restoring to a database using source_app::backup
- Make schedule time non optional

I know we should improve the url stuff, but I think this is good to go

lib/heroku/command/pg_backups.rb
uhoh-itsmaciek: Good call--if you're restoring *from* a Heroku Postgres database to a non-HPG DB, only `name` and `url` fields of the `target` will be valid (it's a `MaybeAttachment`; see below in `resolve_db_or_url`). The

lib/heroku/command/pg_backups.rb
uhoh-itsmaciek: After meatspace convo, that can wait: this should work.

uhoh-itsmaciek:
Yeah, +1. The only thing is the `confirm_command` warning--I think the code I outlined should work for that.

lib/heroku/command/pg_backups.rb
tef: Ah, Nice.

tef:
Added, tested. This is good to go

dickeyxxx:
it's got build failures

tef:
@dickeyxxx I missed a : in the output, specs pass locally now.

PR #1605, opened 2015-06-04
Identifying the cause of duplicate entries in `pg:info`

chadbailey59:
Some customers have seen the same thing I've seen on my own app, where `DATABASE_URL` is set to `HEROKU_POSTGRESQL_COBALT_URL`:

```
~/Code/heroku (cb-database-url) ▸ be bin/heroku pg:info -a cb-rails3
=== DATABASE_URL
Plan:               Standard 0
Status:             Available
Data Size:          7.1 MB
Tables:             1
PG Version:         9.3.5
Connections:        2/120
Fork/Follow:        Available
Rollback:           earliest from 2015-06-04 01:01 UTC
Created:            2015-02-09 21:44 UTC
Forks:              HEROKU_POSTGRESQL_AQUA
Maintenance:        not required
Maintenance window: Thursdays 23:00 to Fridays 03:00 UTC

=== HEROKU_POSTGRESQL_AQUA_URL
Plan:               Standard 0
Status:             Available
Data Size:          7.1 MB
Tables:             0
PG Version:         9.3.6
Connections:        2/120
Fork/Follow:        Available
Rollback:           earliest from 2015-06-04 01:01 UTC
Created:            2015-04-08 16:00 UTC
Forked From:        HEROKU_POSTGRESQL_COBALT
Maintenance:        not required
Maintenance window: Tuesdays 19:30 to 23:30 UTC
...
=== HEROKU_POSTGRESQL_COBALT_URL
Plan:               Standard 0
Status:             Available
Data Size:          7.1 MB
Tables:             1
PG Version:         9.3.5
Connections:        2/120
Fork/Follow:        Available
Rollback:           earliest from 2015-06-04 01:01 UTC
Created:            2015-02-09 21:44 UTC
Forks:              HEROKU_POSTGRESQL_AQUA
Maintenance:        not required
Maintenance window: Thursdays 23:00 to Fridays 03:00 UTC
```

Which is to say, the same database is listed twice in the output: once under COBALT, and once under DATABASE. The expected behavior here is one entry, with a heading of `HEROKU_POSTGRESQL_COBALT_URL (DATABASE_URL)`.

As best I can tell, the dupe is coming back from the API. I was able to trace the difference in the spec and the real output back to the place I inserted those comments; in other words, `@hpg_databases`, which is lightly restructured data straight from the API, includes two different resources in the hash. They have different UUIDs but all the rest of the info is the same.

I updated the spec to try to better represent the data coming back from the API. That said, I have no idea how best to fix it, or else I would have. :)

cc @tef, and anyone else that I should add?

### DIFF
--- a/lib/heroku/helpers/heroku_postgresql.rb
+++ b/lib/heroku/helpers/heroku_postgresql.rb
@@ -109,6 +109,11 @@ module Heroku::Helpers::HerokuPostgresql
       @hpg_databases = Hash[ pairs ]
 
       # TODO: don't bother doing this if DATABASE_URL is already present in hash!
+      # CB: this line is falsing from cb-rails 3 because i do have a DATABASE_URL attachment
+      # so it doesn't resolve to real
+      # Looks like the API is returning the DATABASE_URL as a separate attachment
+      # than the color attachment
+      # the UUIDs are different, but it's the same ["resource"]["id"]
       if !@hpg_databases.key?('DATABASE_URL') && find_database_url_real_attachment
         @hpg_databases['DATABASE_URL'] = find_database_url_real_attachment
       end
--- a/spec/heroku/command/pg_spec.rb
+++ b/spec/heroku/command/pg_spec.rb
@@ -10,7 +10,8 @@ module Heroku::Command
       api.put_config_vars "example", {
         "DATABASE_URL" => "postgres://database_url",
         "HEROKU_POSTGRESQL_IVORY_URL" => "postgres://database_url",
-        "HEROKU_POSTGRESQL_RONIN_URL" => "postgres://ronin_database_url"
+        "HEROKU_POSTGRESQL_RONIN_URL" => "postgres://ronin_database_url",
+        'HEROKU_POSTGRESQL_FOLLOW_URL' => "postgres://follow_database_url"
       }
 
       any_instance_of(Heroku::Helpers::HerokuPostgresql::Resolver) do |pg|
@@ -19,6 +20,13 @@ module Heroku::Command
             'app' => {'name' => 'sushi'},
             'name' => 'HEROKU_POSTGRESQL_IVORY',
             'config_var' => 'HEROKU_POSTGRESQL_IVORY_URL',
+            'resource' => {'name'  => 'loudly-yelling-1232',
+                           'value' => 'postgres://database_url',
+                           'type'  => 'heroku-postgresql:ronin' }}),
+          Heroku::Helpers::HerokuPostgresql::Attachment.new({
+            'app' => {'name' => 'sushi'},
+            'name' => 'DATABASE_URL',
+            'config_var' => 'DATABASE',
             'resource' => {'name'  => 'loudly-yelling-1232',
                            'value' => 'postgres://database_url',
                            'type'  => 'heroku-postgresql:ronin' }}),


lib/heroku/helpers/heroku_postgresql.rb
```diff
@@ -109,6 +109,11 @@ def hpg_databases
       @hpg_databases = Hash[ pairs ]
 
       # TODO: don't bother doing this if DATABASE_URL is already present in hash!
+      # CB: this line is falsing from cb-rails 3 because i do have a DATABASE_URL attachment
+      # so it doesn't resolve to real
+      # Looks like the API is returning the DATABASE_URL as a separate attachment
+      # than the color attachment
+      # the UUIDs are different, but it's the same ["resource"]["id"]
```
bjeanes: This is by design. In the past these attachments were hidden ("implicit attachments") but continuing to do so doesn't make sense in light of shareabale add-ons. i.e. DATABASE shouldn't be treated specially.

bjeanes:
This is probably due to #1564 which specifically stopped hiding DATABASE_URL in response to a bunch of tickets

bjeanes:
> As best I can tell, the dupe is coming back from the API.

It's not a dupe. They are two separate attachmetns to the same database. This command lists databases by attachment. Ideally, we should change it to list by resource instead, as I sketched out in #1564:

```
$ heroku pg:info
=== glistening-goldly-8123
Aliases:     DATABASE, HEROKU_POSTGRESQL_CYAN 
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

=== shining-sweetly-1354
Aliases:     HEROKU_POSTGRESQL_BLACK
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

$ heroku pg:info CYAN
=== glistening-goldly-8123
Aliases:     DATABASE, HEROKU_POSTGRESQL_CYAN 
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

$ heroku pg:info shining-sweetly-1354
=== shining-sweetly-1354
Aliases:     HEROKU_POSTGRESQL_BLACK
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported
```

chadbailey59:
I see what you're saying, Bo. And with all the new attachment stuff, it would make sense to solve this generically, as I've just made it worse on my app:

```
=== Attachments for cb-rails3
Name                        Add-on                  Billing App
--------------------------  ----------------------  -----------
CHAD_CHARCOAL               rolling-freely-9510     cb-rails3
DATABASE                    reading-amply-3184      cb-rails3
DEPLOYHOOKS_HTTP            waking-carefully-7264   cb-rails3
HEROKU_POSTGRESQL_AQUA      hoping-newly-5893       cb-rails3
HEROKU_POSTGRESQL_BLUE      relaxing-slyly-9935     cb-rails3
HEROKU_POSTGRESQL_CHARCOAL  rolling-freely-9510     cb-rails3
HEROKU_POSTGRESQL_COBALT    reading-amply-3184      cb-rails3
HEROKU_POSTGRESQL_PINK      blooming-idly-4485      cb-rails3
HEROKU_POSTGRESQL_YELLOW    cooking-carefully-9671  cb-rails3
```

And sure enough, `CHAD_CHARCOAL` and `HEROKU_POSTGRESQL_CHARCOAL` are the same database, and they're listed twice in `pg:info`.

I'd like @neovintage to comment on this if possible, too. Since pg customers are used to using the attachment names by color already, what if we made one tweak to your proposed output?

```
$ heroku pg:info
=== DATABASE, HEROKU_POSTGRESQL_CYAN
Addon:       glistening-goldly-8123
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

=== HEROKU_POSTGRESQL_BLACK
Addon:       shining-sweetly-1354
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

$ heroku pg:info CYAN
=== DATABASE, HEROKU_POSTGRESQL_CYAN
Addon:       glistening-goldly-8123
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

$ heroku pg:info shining-sweetly-1354
=== HEROKU_POSTGRESQL_BLACK
Addon:       shining-sweetly-1354
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported
```

neovintage:
```
=== HEROKU_POSTGRESQL_CYAN_URL (DATABASE_URL)
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported
Add-on:       glistening-goldly-8123
```

Couple of small tweaks:
* DATABASE_URL is now in parenthesis
* theres a dash between add and on.  This is consistent with `heroku addons -a sushi` output
* Moved addon to the bottom.

bjeanes:
@neovintage so if there were many attachments, are you proposing the header look like:

```
=== HEROKU_POSTGRESQL_CYAN_URL, HEROKU_POSTGRES_MAROON_URL, FOO_URL (DATABASE_URL)
```

I'd also prefer if we can drop the `_URL` if possible. It's not part of the attachment name as technically an add-on can have any number of config vars. But, we can tackle that later if you think it'll cause some friction.

It's worth noting that this still special-cases DATABASE_URL into parenthesis which belies the actual structure, where DATABASE and CYAN are just two aliases. Instead, it indicates that DATABASE is a pointer to CYAN which is an alias to the add-on, which is not the case.

Some installations will *just* have DATABASE so it probably shouldn't show as `(DATABASE)` if it doesn't have another name.

chadbailey59:
We're flirting with a larger conversation here. :) As far as I'm aware, Heroku Postgres is still the only addon *(right now)* in widespread use that has the complexity of multiple addons and attachments. In our view:

* Customers provision addons with unique names like `glistening-goldly-1234`.
* Those addons attach to apps with names like `HEROKU_POSTGRESQL_RED`.
* We create and manage a `URL` config var for each attachment, which ends up being called `HEROKU_POSTGRESQL_RED_URL`.
* We also recommend the convention of using `DATABASE_URL` as the URL of the "primary database" of your application.
* We also provide tools to make that convention easier for customers, such as `heroku pg:promote`.

Hell, I work here, and I'm pretty sure about 95% of that is right. In the customer's view:

* I add a database to my app. It gives me a `DATABASE_URL` as well as this weird `HEROKU_POSTGRESQL_RED_URL` config var.
* I add another database. Now I also have `HEROKU_POSTGRESQL_BLUE_URL`.
* Sometimes when I run `heroku pg` commands, I can just use a color, like `heroku pg:info BLUE`. Sometimes I can run them with no color, and it uses my DATABASE_URL.

I don't think our customers really understand the whole "addons and attachments" thing yet, because they've never had to. We might need to take some baby steps here.

As I've thought about it more, I wonder if we could ease them in with something like:

```
=== HEROKU_POSTGRESQL_CYAN, HEROKU_POSTGRES_MAROON, FOO (DATABASE_URL)
```

Or maybe even have a separate section in the output like:

```
Primary Database (DATABASE_URL): HEROKU_POSTGRESQL_MAROON
You can change your primary database with `heroku pg:promote`.
```

My reasoning behind that is `DATABASE_URL` isn't really an attachment per se; it's a special config variable, but one that is 'locked' to the value of a config variable from a specific attachment. I want customers to understand that addons are different than attachments, and DATABASE_URL is different from both.

neovintage:
Thanks for summarizing chad :-)

I would recommend against showing another section in the output.  The
problem that we're trying to solve is that we want to logically represent
multiple config vars as a single physical database.  Chad, I really like
your suggestion:

```
> As I've thought about it more, I wonder if we could ease them in with
something like:
> === HEROKU_POSTGRESQL_CYAN, HEROKU_POSTGRES_MAROON, FOO (DATABASE_URL)
>
```

bjeanes:
> It gives me a DATABASE_URL as well as this weird HEROKU_POSTGRESQL_RED_URL config var.

That's not true anymore. The V2 API did this, the V3 API does not, as it was a wild special case for HPG. Now, if you install your first DB, you will *just* get `DATABASE`. If you create a new instance and `pg:promote` it, the `pg:promote` command [will create a second attachment for the previous `DATABASE`](https://github.com/heroku/heroku/blob/1816db73f7bd8a9f79119969b21d45430f0b2788/lib/heroku/command/pg.rb#L94) so that it is not orphaned.

> My reasoning behind that is `DATABASE_URL` isn't really an attachment per se; it's a special config variable, but one that is 'locked' to the value of a config variable from a specific attachment. 

Except that it absolutely is an attachment. It used to be just a special var but that caused a whole suite of other problems. It is 100% in every way a non-special attachment (for new installations and new promotions).

> I want customers to understand that addons are different than attachments, and `DATABASE_URL` is different from both.

I want them to understand that `DATABASE_URL` is a variable managed by your database through the attachment named `DATABASE`, because it is.

> I don't think our customers really understand the whole "addons and attachments" thing yet, because they've never had to. We might need to take some baby steps here.

Agreed. So despite my counterpoints, I think your suggested command output is a totally fine interim step. I do think, however, that ultimately it should be resource based. Then longer we misrepresent how add-ons and attachments are, the harder it is for people to start understanding. Right now, the pg commands and web etc are somewhat in dissonance with reality and the understanding we want users to come to. This dissonance was expected and was part of us cutting scope because, as we know, shareable feature set is a big one that had been put off for a long time, and we really wanted to get it done this time!

chadbailey59:
Bo, thanks for laying all this out. I'm happy to be told so clearly how wrong my understanding is! :)

I suspect I'm not the only person that still has the 'old' understanding, and there are a lot of places we should update our documentation to change our terminology. For example:

```
~ ▸ heroku help pg:promote
Usage: heroku pg:promote DATABASE

 sets DATABASE as your DATABASE_URL
```

Should probably become something like:

```
~ ▸ heroku help pg:promote
Usage: heroku pg:promote ADDON_OR_ATTACHMENT

 attaches ADDON_OR_ATTACHMENT as DATABASE, making it your primary database.
```

Likewise, we can reinforce the idea of 'attachments' with something like:

```
Add-on:        glistening-goldly-8123
Attachments:   HEROKU_POSTGRESQL_CYAN, DATABASE
Plan:          Hobby-dev
Status:        Available
Connections:   0/20
PG Version:    9.4.1
Created:       2015-05-13 05:42 UTC
Data Size:     6.4 MB
Tables:        0
Rows:          0/10000 (In compliance)
Fork/Follow:   Unsupported
Rollback:      Unsupported
```

neovintage:
@chadbailey59 what happens to the heading in the db list that you provided?

bjeanes:
@chadbailey59 

> I'm happy to be told so clearly how wrong my understanding is! :)

Haha sorry if it came across as condescension :(

> Likewise, we can reinforce the idea of 'attachments' with something like:

Yeah :+1:

chadbailey59:
@neovintage I think the header goes away, and it's just multiple of these:

```
Add-on:        glistening-goldly-8123
Attachments:   HEROKU_POSTGRESQL_CYAN, DATABASE
Plan:          Hobby-dev
Status:        Available
Connections:   0/20
PG Version:    9.4.1
Created:       2015-05-13 05:42 UTC
Data Size:     6.4 MB
Tables:        0
Rows:          0/10000 (In compliance)
Fork/Follow:   Unsupported
Rollback:      Unsupported

Add-on:        chafing-uncomfortably-1357
Attachments:   HEROKU_POSTGRESQL_ORANGE
Plan:          Hobby-dev
Status:        Available
Connections:   0/20
PG Version:    9.4.1
Created:       2015-05-13 05:42 UTC
Data Size:     6.4 MB
Tables:        0
Rows:          0/10000 (In compliance)
Fork/Follow:   Unsupported
Rollback:      Unsupported
```

The goal being to get people thinking about addons and attachments as 'peers', if you will.

chadbailey59:
@neovintage @bjeanes any thoughts? I know my most recent suggestion will cause a bit of a stir, but I think it helps Bo's message about addons and attachments. I can make the changes if that would help—I just want us to all be on the same page first.

neovintage:
@chadbailey59  What if we use original proposal at the bottom of the comment you made: https://github.com/heroku/heroku/pull/1605#issuecomment-108903496.  What I like about it is that we get rid of the DATABASE_URL in parenthesis and treat it like any other attachment name by including it in a comma separated list.  Also the help text change on `pg:promote` works too.

We will get to resource based naming but, as @bjeanes suggested, let's take a few baby steps to get there.

chadbailey59:
I'm fine with that. For clarity's sake, the new output would look like:

```
$ heroku pg:info
=== DATABASE, HEROKU_POSTGRESQL_CYAN
Addon:       glistening-goldly-8123
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

=== HEROKU_POSTGRESQL_BLACK
Addon:       shining-sweetly-1354
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

$ heroku pg:info CYAN
=== DATABASE, HEROKU_POSTGRESQL_CYAN
Addon:       glistening-goldly-8123
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported

$ heroku pg:info shining-sweetly-1354
=== HEROKU_POSTGRESQL_BLACK
Addon:       shining-sweetly-1354
Plan:        Hobby-dev
Status:      Available
Connections: 0/20
PG Version:  9.4.1
Created:     2015-05-13 05:42 UTC
Data Size:   6.4 MB
Tables:      0
Rows:        0/10000 (In compliance)
Fork/Follow: Unsupported
Rollback:    Unsupported
```

@bjeanes , if you're OK with that, I can probably work on it...

bjeanes:
@chadbailey59 yeah that looks great :+1:

bjeanes:
Resolved in #1644

PR #1612, opened 2015-06-08
Now using HTTPS :sparkles:

rhyselsmore:
Unless there is a clear case where we shouldn't make use of HTTPS, we should use HTTPS.

Historically, the only reason I can think of where HTTP would have been the choice is where there were cert conflicts. @dickeyxxx are you aware of any cases where https didn't work?


dickeyxxx:
everything seems to work still with this!

PR #1613, opened 2015-06-08
HTTPS again :)

rhyselsmore:
Have done a quick flick through the code, this seems to be the last reference.


dickeyxxx:
this one won't work on windows, it can't read s3 assets off of ssl

mathias:
Is there another (trusted) place that we could read VERSION from?

dickeyxxx:
I don't think it's really necessarily personally, this is just an indicator of whether or not an update needs to happen. The download itself is validated against a sha too.

PR #1616, opened 2015-06-14
make ps:scale automatically switch tiers

dickeyxxx:
In general, users shouldn't have to be aware of tiers and just be able
to switch between types no matter what tier the type belongs to.

This makes ps:scale automatically switch between tiers when using
ps:scale to change dyno types.

The one exception to this is that if there are existing dynos the user
isn't transferring to a new type, we don't want to change the tier and
effectively move *all* of their dynos to the lowest level on the new
tier.

This change will check to see if the user is trying to enter a state
like that with some dynos on one tier and some dynos on another and
error out.

Wherever possible the code will simply continue to the API and allow it
to show errors rather than add additional validation on the CLI which
may be too strict.

### DIFF
--- a/lib/heroku/command/ps.rb
+++ b/lib/heroku/command/ps.rb
@@ -17,7 +17,6 @@ class Heroku::Command::Ps < Heroku::Command::Base
   end
   COSTS = Hash[*costs.flatten]
 
-
   # ps:dynos [QTY]
   #
   # DEPRECATED: use `heroku ps:scale dynos=N`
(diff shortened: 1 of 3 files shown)


dickeyxxx:
@brandur @pvh this is pretty well done—though there may be edge cases I haven't thought of. This validation/automation should probably be done in the API but this will at least make the CLI functional with the new types.

dickeyxxx:
not sure if/when you want me to push. I'm not sure what this will do to users not on the new types.

pvh:
probably monday, then?

PR #1622, opened 2015-06-17
Allow user to set --wait-interval

tef:
cf #1603, some customers run multiple copies of pg:backups concurrently. I've used the same option name as in other pg:* commands. /cc @uhoh-itsmaciek


dickeyxxx:
lgtm

PR #1631, opened 2015-06-24
Only split the command by one ':' to separate topic from command.

ys:
This is useful for nested commands.

```
'a:b:c' => { topic: 'a', command: 'b:c' }
```

This seems to be the behavior we get everywhere else. Maybe not a lot of
commands are nested but we start using a second level of nesting.


dickeyxxx:
I tried to support this at every level, I think I just missed it here. Thanks!

ys:
Yes it really looks like a missing one:)

It rocks thanks a lot for the io.js plugin infra, I have fun writing a plugin:D

PR #1651, opened 2015-07-11
Add space option to apps:create

ryanbrainard:
This adds a hidden `--space` option to the `apps:create` command as a first step for moving this off the `dapps` namespace. This still uses a variant until this endpoint is in the mainline API, but we have gotten enough negative feedback about having to use `dapps:create` to create apps in spaces that it is important to at least have this command in the mainline CLI. 

The only other command under `dapps` is `dapps:info`, which will be done separately because 1) it will require API changes and 2) it is less important.

/cc @tim-lang

### DIFF
--- a/lib/heroku/api/organizations_apps_v3_dogwood.rb
+++ b/lib/heroku/api/organizations_apps_v3_dogwood.rb
@@ -1,0 +1,15 @@
+module Heroku
+  class API
+    def post_organizations_app_v3_dogwood(params={})
+      request(
+        :method => :post,
+        :body => Heroku::Helpers.json_encode(params),
+        :expects => 201,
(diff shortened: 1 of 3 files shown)


geemus:
Similar to listing, should there be a short-arg for space here? (this one seems fine, might be good to add that though).

ryanbrainard:
> should there be a short-arg for space here

The option short-arg (`-s`) is already used by `--stack` and didn't really want to add a short args for a hidden option. Perhaps we could use capital `-S` (thoughts?), but again, that's something I'd probably not want to add until this is a visible arg.

geemus:
Not urgent, just seemed inconsistent compared to the other args. Certainly can wait.

PR #1653, opened 2015-07-15
Add space attribute to apps:info

ryanbrainard:
As a follow up to https://github.com/heroku/heroku/pull/1651, this displays the apps' spaces in `apps:info`. This is the last nail in the coffin for the `dapps` namespace. Note, this depends on https://github.com/heroku/api/pull/4464 being deployed; however, the deployment order does not matter since this will only display the attribute if it is present.  

/cc @tim-lang


ryanbrainard:
@dickeyxxx The API change has deployed. This should be good to go.

ryanbrainard:
@dickeyxxx Thanks for the quick merge! Just wanted to make sure you also saw #1651. It would great if these two could go out together.

PR #1655, opened 2015-07-17
Add space option to apps:index

ryanbrainard:
This adds a hidden option to list apps in a given space. 

/cc @dickeyxxx @tim-lang

### DIFF
--- a/lib/heroku/command/apps.rb
+++ b/lib/heroku/command/apps.rb
@@ -9,9 +9,10 @@ class Heroku::Command::Apps < Heroku::Command::Base
   #
   # list your apps
   #
-  # -o, --org ORG  # the org to list the apps for
-  # -A, --all      # list all apps in the org. Not just joined apps
-  # -p, --personal # list apps in personal account when a default org is set
+  # -o, --org ORG     # the org to list the apps for
+  #     --space SPACE # HIDDEN: list apps in a given space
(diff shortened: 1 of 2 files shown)


geemus:
Looks like this wouldn't cope well with specifying an org and a space at the same time presently, thought I would put that on the radar, but doesn't seem like a blocker. Also, should there be a short-arg?

ryanbrainard:
> Looks like this wouldn't cope well with specifying an org and a space at the same time presently

Since spaces are owned by orgs, there should be no need to specify both options. I considered doing the space filtering after the API calls so that it work with the orgs option; however, the v2 org apps endpoint does not include the `space` attribute and adding it gets pretty messy with variants without much benefit. Once `space` is added into mainline in API, this can be simplified and support both options. However, since it would never make sense to specify both options (unless the space is actually owned by the org), it might be better to explicitly block it. What do you think?

> should there be a short-arg?

I didn't add one since there isn't one for `apps:create` (where `-s` would conflict with `--stack`). I also didn't really want to add a short args for a hidden option.

geemus:
Yeah, I didn't think specifying both really made sense, but I wasn't sure. Might be good to explicitly error there.

Fair enough, it just stuck out as being odd/different since the others all appeared to have short versions.

ryanbrainard:
Added explicit error in https://github.com/heroku/heroku/pull/1662

PR #1662, opened 2015-07-21
Explicitly block specifying both --space and --org when listing apps

ryanbrainard:
Follow up to https://github.com/heroku/heroku/pull/1655

/cc @geemus


geemus:
Thanks, might be overkill, but seemed like a quick/easy enough fix.

PR #1663, opened 2015-07-21
Fix issues with backup scheduling

uhoh-itsmaciek:
Previously, if another config var matches the user-provided alias but
its value does *not* match (e.g., the config var for the database the
user is trying to schedule is a substring of another config var with a
different value), we error out. This fixes that and adds specs around
backup scheduling.

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -562,8 +562,10 @@ EOF
     # e.g., names like FOLLOWER_URL work. To do this, we look up the
     # app config vars and re-find one that looks like the user's
     # requested name.
-    db_name, alias_url = resolver.app_config_vars.find { |k,_| k =~ /#{db}/i }
-    if attachment.url != alias_url
+    db_name, alias_url = resolver.app_config_vars.find do |k,v|
+      k =~ /#{db}/i && v == attachment.url
+    end
+    if alias_url.nil?
       error("Could not find database to schedule for backups. Try using its full name.")
     end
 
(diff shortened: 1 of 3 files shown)


lib/heroku/command/pg_backups.rb
```diff
@@ -628,7 +630,7 @@ def hpg_app_client(app_name)
   end
 
   def parse_schedule_time(time_str)
-    hour, tz = time_str.match(/([0-2][0-9]):00 (.*)/) && [ $1, $2 ]
+    hour, tz = time_str.match(/([0-2][0-9]):00 ?(.*)/) && [ $1, $2 ]
```
uhoh-itsmaciek: I also had to make the space here optional because our argument parsing in the helper method seems kinda busted. I'm open to other alternatives if anyone has suggestions.

uhoh-itsmaciek: I think a viable alternative would be using [Shellwords.shellsplit](http://ruby-doc.org/stdlib-2.0.0/libdoc/shellwords/rdoc/Shellwords.html#method-c-shellsplit). I can look into this if that seems like a good idea.

uhoh-itsmaciek: Ok, I tried `Shellwords` and it seems to work, so I changed `spec_helper` to do that. I kept the change making the space optional since having that required seems needlessly pedantic.

spec/heroku/command/pg_backups_spec.rb
```diff
@@ -163,6 +163,63 @@ module Heroku::Command
...
+                                                     'type'  => 'heroku-postgresql:standard-0' }})
+        example_attachments << additional_attachment
+        stub_pg.schedule({ hour: '07', timezone: 'UTC',
+                           schedule_name: 'HEROKU_POSTGRESQL_IVORY_URL' })
+        stderr, stdout = execute("pg:backups schedule HEROKU_POSTGRESQL_IVORY_URL --at '07:00 UTC' --app example")
+        expect(stderr).to be_empty
+        expect(stdout).to match(/Scheduled automatic daily backups/)
+      end
+
+      context "demonstrating cultural imperialism" do
```
tef: :+1:

tef:
This looks good to me. 

Note: The inability to schedule a backup is also holding up particular customers.

PR #1665, opened 2015-07-22
pgbackups: Completed with warnings

tef:
A new state, "Completed with warnings", for when the process is successfully run but pg_restore encounters issues along the way.

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -149,7 +149,12 @@ class Heroku::Command::Pg < Heroku::Command::Base
 
   def transfer_status(t)
     if t[:finished_at] && t[:succeeded]
-      "Finished #{t[:finished_at]}"
+      warnings =  t[:warnings]
+      if warnings && warnings > 0
+       "Completed with #{warnings} warnings" #{t[:finished_at]}"
+      else
+       "Finished #{t[:finished_at]}"
+      end
     elsif t[:finished_at] && !t[:succeeded]
       "Failed #{t[:finished_at]}"
     elsif t[:started_at]
@@ -284,7 +289,12 @@ class Heroku::Command::Pg < Heroku::Command::Base
                client.transfers_get(backup_num, verbose)
              end
     status = if backup[:succeeded]
-               "Completed Successfully"
+               warnings =  backup[:warnings]
+               if warnings && warnings > 0
+                 "Completed with #{warnings} warnings"
+               else
+                 "Completed Successfully"
+               end
              elsif backup[:canceled_at]
                "Canceled"
              elsif backup[:finished_at]
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -151,9 +151,9 @@ class Heroku::Command::Pg < Heroku::Command::Base
     if t[:finished_at] && t[:succeeded]
       warnings =  t[:warnings]
       if warnings && warnings > 0
-       "Completed with #{warnings} warnings" #{t[:finished_at]}"
+       "Finished with #{warnings} warnings" #{t[:finished_at]}"
       else
-       "Finished #{t[:finished_at]}"
+       "Completed #{t[:finished_at]}"
       end
     elsif t[:finished_at] && !t[:succeeded]
       "Failed #{t[:finished_at]}"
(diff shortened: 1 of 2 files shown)


uhoh-itsmaciek:
Mechanically, it looks good. Thinking more about the language, though, I wonder if "Completed with 10 warnings" isn't going to be interpreted as "My backup completed, but there's some stuff that I should look at some point maybe," whereas it could actually mean, "Made some progress but died halfway through." Calling it "Completed" is ambiguous. Maybe "Aborted", but that brings to mind an aborted transaction, which does roll back all changes, which is the other ambiguity we are trying to avoid in the first place. "*&^# the bed"? Perhaps a little too colloquial. I spent a few minutes at http://thesaurus.com but couldn't come up with anything better, so maybe we should just stick with "Completed with warnings," but I hope we can come up with something clearer.

Thougths @chadbailey59 @neovintage?

tef:
We have four states for when a backup/restore/transfer ends:

Finished (finished_at not nil, successful is true, warnings is nil/0)
Completed (finished_at not nil, successful is true, warnings is nonzero)
Failed (finished_at not nil, successful is false, cancelled_at is nil)
Cancelled (finished_at not nil, successful is false, cancelled_at not nil)

The latter two mean that the processes terminated early or unexpectedly, and cancelled means it was due to user action, rather than software failure. The first two mean that pg_restore ran and exited, either without errors, or with some known errors on restore (extension comments).

@uhoh-itsmaciek: the language is awful, yep, but I am struggling to find something better.

Maybe we should have "Finished without problems" "Finished with N problems" "Failed" Cancelled"

neovintage:
@uhoh-itsmaciek @tef are the warnings that we surface actionable for the customer?

uhoh-itsmaciek:
@neovintage Sometimes? The warnings are anything that `pg_restore` could "warn" about, which is basically the failure of any individual statement in the restore script.

neovintage:
I think @tef is onto something with the states. Instead of using the word "problem", I would still use "warning". Warning means that it could be ignored if you want to. Problem connotes that I have to do something. Based on the states this my suggestion:

* Finished
* Finished (5 warnings)
* Failed
* User Cancelled

The "finished" means that it completed successfully with no problems.  The second, finished with warnings, gives customers something to check into if they want. Failed is just that. I put user in front of cancelled to make it clear that it wasn't the system that did it.

feedback?  /cc @uhoh-itsmaciek @tef

uhoh-itsmaciek:
I think "warnings" is the wrong word: the restore could be utterly useless and we'll say "Finished (3 warnings)." I prefer tef's "problems," but maybe there's something better still? Basically, the issue is "Finished with warnings" is semantically closer to "Failed," but the new nomenclature sweeps that under the rug. In fact, right now they are the same thing--we're putting in the distinction because @chadbailey59 mentioned users sometimes get confused when a failed restore still creates some tables in their db. But "Finished with warnings" aligns it with "Finished," which seems misleading.  Maybe just "Failed (partial restore)" or something along those lines?

chadbailey59:
> Thinking more about the language, though, I wonder if "Completed with 10 warnings" isn't going to be interpreted as "My backup completed, but there's some stuff that I should look at some point maybe," whereas it could actually mean, "Made some progress but died halfway through."

I've been presuming these are all cases of the former, not the latter. Won't pg_restore error out if something truly terrible happens? The case we're trying to cover for here is when pg_restore thinks it completed successfully while ignoring a few """minor""" errors. pg_restore seemed to determine they were OK to ignore, so after several laps in my head, I *think* i'm in favor of tef's original wording of "Finished" and "warnings".

uhoh-itsmaciek:
@chadbailey59 no, not really: `pg_restore` offers no distinction between "something truly terrible" and "minor" errors. We could try to determine this heuristically, but that way lies madness.

neovintage:
If we can't make a distinction between errors that need a customer's attention (terrible) and those that don't (minor), I'm inclined to err on the side of stronger language that makes the customer look into it and separate it from a backup that finished with no problems. Although, we'd need to get ahead of the errors by documenting them in the devcenter so that customers know if they should do something about it:

* completed
* finished with problems
* failed 
* user cancelled

chadbailey59:
/me shakes his fist at pg_backups

I like Rimas's most recent list. Using "Completed" when we know it was good, but a more ambiguous "Finished" when we can't be sure.

tef:
Ok, will change this to rimas' suggestion and update PR, then hopefully merge

PR #1667, opened 2015-07-22
Fix spaces with default org

ryanbrainard:
This fixes the new apps's spaces commands to work with the `HEROKU_ORGANIZATION` env to set default org. The changes are as follows:

 - Users can explicitly specify `--space` or `--org`, but not both.
 - If the `HEROKU_ORGANIZATION` is set, it is ignored when using `--space`. The reason being that `--org` overrides `HEROKU_ORGANIZATION`, so `--space` (which is in an org) should too. In the future we might want to do something fancy where we look up the org from the space to reconcile these two, but since space names are globally unique, there's probably little reason users would want that.
 - Added `shared_examples` tests with and without a default org for app space commands. 

/cc @geemus @markpundsack

### DIFF
--- a/lib/heroku/command/apps.rb
+++ b/lib/heroku/command/apps.rb
@@ -27,11 +27,8 @@ class Heroku::Command::Apps < Heroku::Command::Base
   #
   def index
     validate_arguments!
+    validate_space_xor_org!
     options[:ignore_no_org] = true
-
-    if options[:space] && org
-      error "Specify option for space or org, but not both."
-    end
 
     apps = if options[:space]
       api.get_apps.body.select do |app|
@@ -44,7 +41,10 @@ class Heroku::Command::Apps < Heroku::Command::Base
     end
 
     unless apps.empty?
-      if org
+      if options[:space]
+        styled_header("Apps in space #{options[:space]}")
+        styled_array(apps.map { |app| regionized_app_name(app) })
+      elsif org
         joined, unjoined = apps.partition { |app| app['joined'] == true }
 
         styled_header("Apps joined in organization #{org}")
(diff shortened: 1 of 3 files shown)


ryanbrainard:
Closing this for the time being, as there is a better way this can potentially work I want to try.

ryanbrainard:
Tried my experiment, but it will require API changes. Re-opening as-is.

geemus:
@ryanbrainard could you elaborate a bit on why you dislike this and what the alternate might be? Just want to ensure I have proper context. Thanks!

ryanbrainard:
@geemus There is a separate issue where unjoined apps are not showing up for `heroku info --space SPACE`, and I was hoping to kill two birds with one stone with a slightly different implementation. The problem with the unjoined apps is that we're using `api.get_apps` instead of `org_api.get_apps(org)`. If we want to use `org_api.get_apps(org)`, we will need to look up the org from the space client side (do-able) and also make an API change to include the `space` in the serialization of `GET /v1/organization/:org/app`. I didn't make that serialization change previously because v3 uses the same serialization, which would be a change to a production-level endpoint. Alternatives would be either 1) have separate serializers with a variant and/or 2) introduce a new endpoint to list apps in a space (which we punted on before since it wasn't clear how the URL should be formatted). I'd be happy hear your thoughts on this.

For the time being though, this change should be fine for fixing the issue at hand (i.e. handling `HEROKU_ORGANIZATION`). I was just hoping I could handle the two together.

geemus:
Got it. Sounds like a case where filters would be helpful to avoid some of the roundtrips, though I don't think we are quite prepared to use them.

geemus:
Does this need another release?

ryanbrainard:
@geemus Yes, you can release it at your convenience.

geemus:
Released in 3.40.6

ryanbrainard:
Thanks Wes!

PR #1678, opened 2015-07-30
Use resource name to resolve services.

cyberdelia:
This solves problems with linking shared addons.

### DIFF
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -554,12 +554,12 @@ class Heroku::Command::Pg < Heroku::Command::Base
       error("Usage links <LOCAL> <REMOTE>") unless [local, remote].all?
 
       local_attachment = generate_resolver.resolve(local, "DATABASE_URL")
-      remote_attachment = resolve_service_or_url(remote)
+      remote_attachment = resolve_service(remote)
 
       output_with_bang("No source database specified.") unless local_attachment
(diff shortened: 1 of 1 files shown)


cyberdelia:
cc/ @fnando

fnando:
:+1:

cyberdelia:
Actually, there is a case that this doesn't handle well.

cyberdelia:
There you go, fixed.

lib/heroku/command/pg.rb
```diff
@@ -588,21 +588,13 @@ def links
...
+    attachment = (resolve_addon(name) || []).first
 
-      error("Remote database could not be found.") unless attachment
-      error("Remote database is invalid.") unless attachment['addon_service']['name'] =~ /heroku-(redis|postgresql)/
+    error("Remote database could not be found.") unless attachment
+    error("Remote database is invalid.") unless attachment['addon_service']['name'] =~ /heroku-(redis|postgresql)/
 
-      MaybeAttachment.new(attachment_name, get_config_var(attachment['config_vars'].first), attachment)
-    end
+    MaybeAttachment.new(attachment['name'], nil, attachment)
```
bjeanes: `resolve_addon` doesn't return an attachment, though, it returns an add-on resource. I'm confused what this is doing. Is there code here assuming that this will be an attachment?

PR #1680, opened 2015-08-03
added cli analytics

dickeyxxx:


lib/heroku/analytics.rb
```diff
@@ -0,0 +1,46 @@
...
+    fork do
+      commands = lines.map do |line|
+        line = line.split('|')
+        {command: line[0], timestamp: line[1].to_i}
+      end
+      payload = {
+        user:     user,
+        commands: commands
+      }
+      Excon.post('https://heroku-cli-analytics.herokuapp.com/record', body: JSON.dump(payload))
```
markpundsack: Should we use a heroku.com domain for this?

jacobian:
A couple of comments from my security POV:

- Out of respect for user privacy -- and to comply with Safe Harbor -- this needs to be something users can opt out of. Ideally, this would be opt-in: ask users the first time this runs if they're OK sharing this data, and only send it if they consent. Opt-out might be acceptable if there's a reason why opt-in won't work, but as a general principle asking before "phoning home" is a good privacy practice, and we should follow it if we can.

- As @markpundsack mentions, this should be a `heroku.com` domain, to make it clear it's first-party.

markpundsack:
Agreed on being good policy to ask up-front, but technically I thought we only *needed* to provide out-out for third-party analytics. If this is going directly to us, I was under the impression we didn’t absolutely have to provide an opt-out. Now my knowledge might be out of date, and certainly predates us being Safe Harbor, but when we implemented this in Dashboard, we made the opt-out explicitly for third-party. I don’t think we actually have any first-party analytics in Dashboard, so I don’t think we’re in violation either way. But it would be good to get clarity.

image

dickeyxxx:
thanks for looking @jacobian! We can get opt-in functionality in, it's a little tricky since I need to make it some kind of auto-canceling opt-in (in case an automated script is running the CLI), but it's doable.

dickeyxxx:
The prompting + timeout is now in. Just waiting this before we can release https://github.com/heroku/foundation/issues/826

lib/heroku/analytics.rb
```diff
@@ -0,0 +1,64 @@
...
+      f.write("#{command}|#{Time.now.to_i}\n")
+    end
+  rescue
+  end
+
+  def self.submit
+    return if skip_analytics
+    lines = File.read(path).split("\n")
+    return if lines.count < 10 # only submit if we have 10 entries to send
+    fork do
```
dickeyxxx: hmm it seems this won't work on windows

PR #1681, opened 2015-08-03
ensure that HEROKU_HEADERS are used for pg:backups commands

dickeyxxx:


chadbailey59:
@tef are you OK with merging this? I'd like to get this fixed so I can get back to sudo-ing properly :)

dickeyxxx:
I think I intended to merge this before actually

PR #1691, opened 2015-08-17
Fix `heroku apps --space` to display unjoined apps

ryanbrainard:
This changes `heroku apps --space` to use the updated v2 org apps endpoint that now includes the space attribute. Unjoined apps in spaces are now displayed and the rest of the space query and display logic falls inline with orgs now. This also has the side benefit of validating space visibility to improve error messages for unauthorized access.

@geemus This fixes the issue I was talking about in https://github.com/heroku/heroku/pull/1667#issuecomment-124211467. I went with alternative 1 in API, and this is the corresponding CLI change. It unifies the space and org logic without introducing any new endpoints, so I think this is a win.

/cc @dickeyxxx

### DIFF
--- a/lib/heroku/api/spaces_v3_dogwood.rb
+++ b/lib/heroku/api/spaces_v3_dogwood.rb
@@ -1,0 +1,14 @@
+module Heroku
+  class API
+    def get_space_v3_dogwood(space_identity)
+      request(
+        :method => :get,
+        :expects => [200],
+        :path => "/spaces/#{space_identity}",
+        :headers => {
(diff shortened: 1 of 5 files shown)


geemus:
Looks fine to me, but I think @dickeyxxx should be back soon, so will probably leave it for his scrutiny.

ryanbrainard:
I just opened https://github.com/heroku/heroku/pull/1692, which layers on this pull. I did it that way instead of basing it directly on `master` to avoid merge conflicts. This can go in without it, but it would nice for them to get merged together.

PR #1692, opened 2015-08-18
Always display unjoined org apps

ryanbrainard:
**Note, the base for this PR is `space-apps-unjoined` for a clearer diff of the what's changing and to avoid a future merge conflict; however, it should be merged all the way to `master`.**

This changes the org app list to always display joined and unjoined apps together in one list for parity with Dashboard. Hiding and segregating unjoined apps has proved to confuse users.

This leaves the --all option in place, but updates the documentation to reflect what is actually does for the personal app list.

/cc @naaman @dickeyxxx

### DIFF
--- a/lib/heroku/command/apps.rb
+++ b/lib/heroku/command/apps.rb
@@ -12,7 +12,7 @@ class Heroku::Command::Apps < Heroku::Command::Base
   #
   # -o, --org ORG     # the org to list the apps for
   #     --space SPACE # HIDDEN: list apps in a given space
-  # -A, --all         # list all apps in the org. Not just joined apps
+  # -A, --all         # list all collaborated apps, including joined org apps in personal app list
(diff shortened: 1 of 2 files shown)


naaman:
Seems like we should just get rid of `--all`; it's more confusing with this change than helpful. I'm not sure why we'd filter out org apps at this point.

Aside from introducing a UX issue, :+1:.

ryanbrainard:
@naaman I went back and forth on that, but decided to leave it because 1) it makes it act like the personal app list on Dashboard, 2) it scopes the change to the org app list only, 3) does not break scripts that use the flag today. If we do completely remove `--all`, we would need to decide what the behavior would be for the personal app list. If we want to make it like Dashboard, the list would to be never include org apps (even joined apps) in the personal app list.

PR #1695, opened 2015-08-21
support custom kernels

curzonj:
@dickeyxxx As this doesn't pose a security issue (it's a protected ability) can we include this?


dickeyxxx:
sure!

PR #1705, opened 2015-08-28
Correctly fallback on service:plan add-on resolution

bjeanes:
`heroku addons:open -a APP ADDON_SERVICE` was failing because it would first look up `ADDON_SERVICE` as an add-on name. heroku/api#4725 meant that this endpoint returned something other than a 404 for a string like `papertrail`. This accounts for that. Fortunately heroku/api#4725 actually means this entire resolver can be simplified drastically as most of this resolution logic now lives in the API. I'll handle that in a separate PR, though, as it'll take more time.

cc @heroku/add-ons @dickeyxxx

### DIFF
--- a/lib/heroku/helpers/addons/resolve.rb
+++ b/lib/heroku/helpers/addons/resolve.rb
@@ -10,9 +10,6 @@ module Heroku::Helpers
       RESOURCE     = /^@?([a-z][a-z0-9-]+)$/
       SERVICE_PLAN = /^(?:([a-z0-9_-]+):)?([a-z0-9_-]+)$/ # service / service:plan
 
-      class AddonDoesNotExistError < Heroku::API::Errors::Error
-      end
-
       # Finds attachments that match provided identifier.
       #
       # Always returns an Array of 0 or more results.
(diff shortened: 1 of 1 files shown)


lib/heroku/helpers/addons/resolve.rb
```diff
@@ -85,16 +82,23 @@ def resolve_addon(identifier, &filter)
...
             rescue Heroku::API::Errors::Forbidden
               # treat permission error as no match because there might exist a
               # resource on someone else's app that has a name which
               # corresponds to a service name that we wish to check below (e.g.
               # "memcachier")
+            rescue Heroku::API::Errors::RequestFailed => e
+              error_id = (json_decode(e.response.body.to_s) || {})["id"]
+              if error_id == "multiple_matches"
+                # ignore; name-based matches are never ambiguous, so this
+                # identifier is probably a service/plan
```
mnoble: For the `ADDON_SERVICE` scenario, isn't the above only true when the request is scoped to an app? If they did `heroku addons:open papertrail` in a non-app directory, don't we still want the "Ambiguous identifier" error?

bjeanes: Yeah it's tricky. The next piece will prompt for `app` which will error and ask for `--app`, so I think it always ends up in the app-scoped case when using those identifiers. We *could* have it be an ambiguous match for all the `papertrail`s across all app, but the way to handle that will be to ask for an app anyway, so it's really just moving the error. In any case, #1706 should make all of this more sane. What do you think?

bjeanes:
@dickeyxxx bump! getting a few tickets about this.

lib/heroku/helpers/addons/resolve.rb
mattreduce: A-ha, thanks for explaining. That should do for now (and #1706 will help when it lands) :+1:

mattreduce:
Looks good to me, thanks @bjeanes!

PR #1706, opened 2015-08-28
Lean on API for add-on resolution

bjeanes:
Yay heroku/api#4725!

I'm expecting a few test failures here that I'll need to fix. Furthermore, this should have a fair amount of manual QA done (cc @heroku/add-ons!) before this is merged.

### DIFF
--- a/lib/heroku/command/addons.rb
+++ b/lib/heroku/command/addons.rb
@@ -352,26 +352,13 @@ module Heroku::Command
 
       # If it looks like a plan, optimistically open docs, otherwise try to
       # lookup a corresponding add-on and open the docs for its service.
-      if identifier.include?(':')
-        service = identifier.split(':')[0]
-        launchy("Opening #{service} docs", addon_docs_url(service))
+      if identifier.include?(':') || get_service(identifier)
+        launchy("Opening #{identifier} docs", addon_docs_url(identifier))
       else
         # searching by any number of things
-        matches = resolve_addon(identifier)
-        services = matches.map { |m| m['addon_service']['name'] }.uniq
-
-        case services.count
-        when 0
-          # Optimistically open docs for whatever they passed in
-          launchy("Opening #{identifier} docs", addon_docs_url(identifier))
-        when 1
-          service = services.first
-          launchy("Opening #{service} docs", addon_docs_url(service))
-        else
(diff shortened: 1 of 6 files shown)


mathias:
This works well enough for me locally. But I don't really have very creative use cases in attempting to break it.

Maybe we can write tests (or at least manually reproduce and validate)  some of the more complex cases we heard from Postgres team for things like fork, follow, etc where the resolution was an issue?

bjeanes:
@mathias all those kinds of tests [exist in API](https://github.com/heroku/api/blob/591ba59c50bce7f2ff465241b7acfed5a4fd371a/spec/mediators/addons/resolver_spec.rb) already. Since this is just an API call and not much else, I avoided adding a lot of tests that were just hitting Excon stubs which may or may not actually reflect API behaviour. This feels like something that is better tested in an integration fashion, which we don't have an answer for in the toolbelt, so thus QA. Do you have an idea of a kind of test we could add that differs from my assumptions? Would definitely be interested in exploring, if so...

mathias:
Not really, but I didn't understand the transfer of responsibility up to the API, I guess. In that case, tests need only validate that for various types of arguments which might be treated differently (And really, I'm guessing they're not since the CLI doesn't care whether a string looks like "DATABASE" or "violet" or "papertrail") and that they all respond with the same stub, then we're good on testing it. There's no point in trying to assert anything here when the client is becoming thinner in a "thin client" sense.

mattreduce:
@bjeanes @mathias I think [Direwolf](https://github.com/heroku/direwolf) is a good place to test various scenarios for the CLI. Thoughts?

mathias:
That's a good point. That is the kind of integration thing we'd want to see work from both sides interacting, so it's in Direwolf's wheelhouse. 

I don't want to step on toes, but if I can help in writing Direwolf tests, let me know.

bjeanes:
Direwolf to test the integration makes a lot of sense. As I said in stand up, I'm mostly concerned about the invocation styles I might not know about, but Direwolf tests for those that we do seems valuable.

bjeanes:
@dickeyxxx I think this is probably good for merge.

dickeyxxx:
not till next week

bjeanes:
@dickeyxxx of course

PR #1707, opened 2015-08-28
Use same attachment for multiple invocations of exec_sql

gregburek:
Running pg:ps will execute two sql statements, one to determine the
database version and another to read pg_stat_activity. Two executions of
exec_sql will shift the supplied argument out, which leaves the defaul,
DATABASE_URL. This commit uses a cached class variable so that multiple
exec_sql calls will run on the first supplied database attachment.


uhoh-itsmaciek:
So `exec_sql` has pretty awful semantics here, but fixing that would be significantly more invasive, so +:elephant:...

PR #1708, opened 2015-08-31
Fixed bug in setting buildpack at end of list

jkutner:


freeformz:
@jkutner Since these tests are stubbing put are these output assertions even testing anything?

jkutner:
@freeformz yea, the tests are missing a decent chunk of the flow. But in this case the test actually does capture the bug.

freeformz:
hunh. Well then :+1: then.

PR #1713, opened 2015-09-03
Add option to copy config variables between apps

gerep:
When migrating an Application to another stack, Heroku's help page suggests to create a clone of that application in the new stack and test it. It is also required to copy the variables from that application.

Command:

`heroku config:copy target_app --app source_app`

### DIFF
--- a/lib/heroku/command/config.rb
+++ b/lib/heroku/command/config.rb
@@ -156,4 +156,25 @@ class Heroku::Command::Config < Heroku::Command::Base
 
   alias_command "config:remove", "config:unset"
 
+  # config:copy APP2
+  #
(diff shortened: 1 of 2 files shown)


gerep:
How is it that with a different Ruby version the tests are failing this way?

gerep:
Anyone?

dickeyxxx:
apologies since this isn't really mentioned anywhere you would've likely seen it, but we're trying not to add any new functionality to this version of the toolbelt. We're working on the replacement here: https://github.com/heroku/heroku-cli

gerep:
@dickeyxxx OK, got it, thanks for the heads up!

PR #1714, opened 2015-09-04
added support for CET and CEST

mandagill:
We tend to see a lot of tickets from folks trying to schedule Transferatu backups using CET and CEST. Adding the ability to parse this so they can schedule the backup using this option.


tef:
+1. @dickeyxxx this seems good to merge whenever (we don't have any open issues for it)

PR #1730, opened 2015-09-16
pg:* add-on resolve with API

bjeanes:
Similar to #1706, this teaches the `pg:*` commands to resolve add-on identifiers with the API when no obvious local match is found locally. This is a minimal change because the current code uses APIv2 extensively and a lot of commands rely on it. It seemed the most prudent thing, for now, to just add an additional look up instead of rewriting the whole lookup logic.

cc @heroku/add-ons @dickeyxxx


mnoble:
Hooray for consolidating logic!

bjeanes:
bump @dickeyxxx

dickeyxxx:
so ultimately does this perform a second http request if the match isn't found? I'm a bit concerned how this would work with paranoid apps

bjeanes:
Yes it does, sometimes even 2. I think the ultimate answer here is to port these commands to JS plugins, but that's a lot of work out of the gate. Though I did briefly discuss it with @uhoh-itsmaciek and @kvlou yesterday. Is there a way to address the paranoid concern safely?

dickeyxxx:
any command that uses these will need to have `requires_preauth` it's hard to tell which ones though

garyposter:
@bjeanes @mattreduce @dickeyxxx is the 0.08% drop in coverage enough to block this?  If so, could one of you top that up quickly?  I'd really love to see this merged and closed ASAP.  Thank you!

mathias:
I'm looking at the coverage now and trying to orient myself, will defer to @mattreduce if he's got a quicker solution.

dickeyxxx:
I don't care about the coverage. I'm out at lunch with a friend but I can release this after.

dickeyxxx:
There are merge conflicts though

mathias:
Resolved conflicts, @dickeyxxx can you confirm that these files are supposed to be deleted? Seeing this many things disappear on a merge was worrying:
lib/heroku/command/two_factor.rb lib/heroku/command/status.rb lib/heroku/command/run.rb lib/heroku/command/maintenance.rb lib/heroku/command/local.rb lib/heroku/command/git.rb lib/heroku/command/fork.rb lib/heroku/command/apps.rb (apps:info commad, not the whole file)

dickeyxxx:
yes since they all moved into v4 plugis

garyposter:
Awesome thank you @mathias and @dickeyxxx !

mattreduce:
:tada:

PR #1734, opened 2015-09-18
Consolidate add on related output

mnoble:
The pg:info command has been updated to display attachments and resources in a consolidated way that makes more sense in a Shareable world.

All this really means is not displaying multiple groups of data, if they are for the same logical database as well as displaying the Resource name in addition to Attachment names.

Before:

```
=== HEROKU_POSTGRESQL_IVORY_URL # resource: walking-slowly-1234
No data sources are linked into this database.

=== HEROKU_POSTGRESQL_RED_URL # resource: walking-slowly-1234
No data sources are linked into this database.
```

After:

```
=== HEROKU_POSTGRESQL_RED_URL, HEROKU_POSTGRESQL_IVORY_URL (walking-slowly-1234)
No data sources are linked into this database.
```

This also makes a couple minor things like indentation and capitalization consistent within PG commands.

@dickeyxxx 
@heroku/department-of-data 
cc @heroku/add-ons

### DIFF
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -42,9 +42,9 @@ class Heroku::Command::Pg < Heroku::Command::Base
 
   # pg:info [DATABASE]
   #
-  #   -x, --extended  # Show extended information
-  #
   # display database information
+  #
+  #   -x, --extended  # Show extended information
   #
   # If DATABASE is not specified, displays all databases
   #
@@ -118,9 +118,9 @@ class Heroku::Command::Pg < Heroku::Command::Base
 
   # pg:psql [DATABASE]
   #
-  #  -c, --command COMMAND      # optional SQL command to run
-  #
   # open a psql shell to the database
+  #
+  #  -c, --command COMMAND      # optional SQL command to run
   #
   # defaults to DATABASE_URL databases if no DATABASE is specified
   #
@@ -501,7 +501,7 @@ class Heroku::Command::Pg < Heroku::Command::Base
 
   # pg:links <create|destroy>
   #
-  # Create links between data stores.  Without a subcommand, it lists all
+  # create links between data stores.  Without a subcommand, it lists all
   # databases and information on the link.
   #
   # create <REMOTE> <LOCAL>   # Create a data link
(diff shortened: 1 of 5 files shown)


lib/heroku/command/pg.rb
```diff
@@ -527,13 +527,15 @@ def links
...
       error("No database attached to this app.") if dbs.compact.empty?
 
-      dbs.each_with_index do |attachment, index|
-        response = hpg_client(attachment).link_list
+      dbs_by_addons.each_with_index do |(resource, attachments), index|
+        response = hpg_client(attachments.first).link_list
         display "\n" if index.nonzero?
 
-        styled_header("#{attachment.display_name} (#{attachment.resource_name})")
+        styled_header("#{attachments.map(&:config_var).join(", ")} (#{resource})")
```
cyberdelia: I did the reverse in the redis plugin: ``resource (config_vars, ...)``

mnoble: I was following what Bo was doing in #1730. We should probably come up with some sort of style-guide for stuff like this.

mattreduce: fwiw I prefer the `resource (config_vars, ...)` format

mnoble: In terms of precedence for add-ons, I agree, but there's a long history in HPG of attachments being the first-class identifier.

mattreduce: Indeed, and I'd like to see that change :speak_no_evil: Perhaps you're right that we shouldn't surprise users, though. The struggle is real. :weary:

lib/heroku/command/pg_backups.rb
```diff
@@ -8,7 +8,7 @@ class Heroku::Command::Pg < Heroku::Command::Base
   #
   #   --wait-interval SECONDS      # how frequently to poll (to avoid rate-limiting)
   #
-  # Copy all data from source database to target. At least one of
+  # copy all data from source database to target. At least one of
```
dickeyxxx: +1 on keeping the consistency here

lib/heroku/command/pg.rb
dickeyxxx: cc @mattgraham

PR #1738, opened 2015-09-21
Add a pg:tunnel command for opening SSH tunnels

gulopine:
This allows users to access private VPC databases from the outside, using non-psql tools, including GUIs. The command simply opens an SSH tunnel and outputs the local port to use for the Postgres connection. Also includes an optional parameter for specifying a port number, which will allow users to store a connection config in their tools of choice, using a consistent port number.

I couldn't find any prior tests for the bastion tunneling, so I haven't added any here either. Any guidance in that area would be appreciated, if such tests are necessary for a merge.

### DIFF
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -164,6 +164,29 @@ class Heroku::Command::Pg < Heroku::Command::Base
     end
   end
 
+  # pg:tunnel [DATABASE]
+  #
+  # open an SSH tunnel to a database
+  #
+  # defaults to DATABASE_URL databases if no DATABASE is specified
+  #
+  def tunnel
+    requires_preauth
+    attachment = generate_resolver.resolve(shift_argument, "DATABASE_URL")
+    validate_arguments!
+
+    uri = URI.parse( attachment.url )
+
+    if attachment.uses_bastion?
+      attachment.maybe_tunnel do |uri|
+        puts "---> Tunnel to #{attachment.display_name} open at #{uri.host}:#{uri.port}"
+        sleep
+      end
+    else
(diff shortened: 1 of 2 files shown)


fdr:
Do you got any good idea for testing? I didn't. It's probably worth somehow stubbing out `maybe_tunnel` and seeing if one can test the other behavior properly. I didn't add any new commands, so the user-apparent surface area was equivalent, whereas this adds a new one.

If you don't want to launch this directly into the main toolbelt, it can also live in https://github.com/heroku/heroku-pg-extras/ with a lower bar of quality and necessity. That's how commands tend to get experimented with and promoted to toolbelt after being stable and useful for a time.

dickeyxxx:
The ruby code is getting close to deprecation and I'd really rather not add more commands to it, if this could be a node plugin that would be much more desirable.

gulopine:
I had wondered about the Node side of things, but I didn't see any prior art there for the tunneling behavior, so I started here. pg-extras does seem like a better place for this to live, though it doesn't look like there's been much activity over there lately. I'm happy to drop this PR and move it to pg-extras until there's a clear place to put a Node version of it instead.

dickeyxxx:
We need a plugin to put the node code for pg stuff. I could at least get that started if you like.

fdr:
On Wed, Oct 14, 2015 at 9:03 AM Jeff Dickey <notifications@github.com>
wrote:

> We need a plugin to put the node code for pg stuff. I could at least get
> that started if you like.
>

Did you know Tim implemented an apparently functional "maybeTunnel" for
node in the Redis plugin? Maybe this can be of some interest. Though Marty
also did quite a bit of Node SSH tunneling work, so he's pretty familiar
with those mechanics as well...

PR #1762, opened 2015-10-05
Improve handling of shareable add-ons w.r.t. Postgres backups

uhoh-itsmaciek:
Since Postgres backups are tied to the app and not the add-on, but
we require *an* add-on to interact with, make sure that that add-on
is attached to the right app.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -100,6 +100,3 @@ DEPENDENCIES
   rr
   rspec
   webmock
-
-BUNDLED WITH
-   1.10.6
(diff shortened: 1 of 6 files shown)


uhoh-itsmaciek:
No specs yet, but I'll follow up and add these. Basically, this fixes an issue that can manifest in some backups commands running (or most likely attempting to run) against the wrong app because of how attachable resources changed some assumptions in the Postgres resolver code.

uhoh-itsmaciek:
Ok, fixed specs and added a new one testing the issue.

PR #1777, opened 2015-10-13
Discontinue use of v3.domain-cname variant

ryanbrainard:
This removes the use of the merged v3.domain-cname API variant. It also extracts the v3 media types in the `Heroku::API` class to constants.

cc: @dickeyxxx @jsullivan


dickeyxxx:
can you take out the constants? I'm not a fan

ryanbrainard:
@dickeyxxx Just removed the constants.

jsullivan:
LGTM

dickeyxxx:
Thanks!

PR #1805, opened 2015-11-04
update v4 plugins when running plugins:update

dickeyxxx:


dickeyxxx:
fyi @keiko713

keiko713:
wow thank YOU! this'll save my life :heart_eyes:

PR #1816, opened 2015-11-13
Use attachment SSO url if it can be determined

bjeanes:
This is a bit finicky compared to the Dashboard because we're getting an
arbitrary identifier which might be for an add-on or an attachments. This
attempts to resolve an attachment with the identifier but silently falls back
on add-on if it gives either a 404 or a 422 for multiple matches.

attn: @heroku/add-ons

### DIFF
--- a/spec/heroku/command/addons_spec.rb
+++ b/spec/heroku/command/addons_spec.rb
@@ -771,7 +771,52 @@ STDERR
       end
 
       it "opens the addon if only one matches" do
+        Excon.stub(method: :get, path: %r'(/apps/example)?/addon-attachments/redistogo:nano') do
+          {status: 404}
+        end
         addon.merge!(addon_service: { name: "redistogo:nano" })
+
+        allow_any_instance_of(Heroku::Command::Addons).to receive(:resolve_addon!).and_return(stringify(addon))
+        require("launchy")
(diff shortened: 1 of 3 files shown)


spec/heroku/command/addons_spec.rb
```diff
@@ -771,7 +771,52 @@ module Heroku::Command
       end
 
       it "opens the addon if only one matches" do
+        Excon.stub(method: :get, path: %r'(/apps/example)?/addon-attachments/redistogo:nano') do
+          {status: 404}
+        end
         addon.merge!(addon_service: { name: "redistogo:nano" })
+
+        allow_any_instance_of(Heroku::Command::Addons).to receive(:resolve_addon!).and_return(stringify(addon))
+        require("launchy")
```
mnoble: Is `launchy` not required in the code?

mnoble:
Looks good to me. :+1:

mattreduce:
Nice! Thank you, @bjeanes :sparkles:

mattreduce:
@dickeyxxx my apologies for the hasty merge here—when I saw that Bo's code looked right to me and moved back to the discussion, I was in the mindset that I was working with our CLI plugin. Feel free to un-merge this until you've had a chance to review it.

spec/heroku/command/addons_spec.rb
dickeyxxx: only when needed to make the CLI boot faster

dickeyxxx:
lgtm

PR #1817, opened 2015-11-13
Fixing tests so they can run under windows

ransombriggs:
@dickeyxxx review?  These are only test changes, I was working on fixing a windows bug and found that the test suite was not passing, so I fixed the tests.


dickeyxxx:
lgtm

PR #1822, opened 2015-11-17
spaces: don't use platform api dogwood variant

jsullivan:
The platform api dogwood variant is merging into prod v3. We don't need to make calls to the platform api requesting the variant for dogwood-specific properties of organizations, regions, or spaces. Calls to prod v3 will return the same properties now.

*Note: please don't merge and deploy these changes until heroku/api#5162 has been merged and deployed.

### DIFF
--- a/lib/heroku/api/organizations_apps.rb
+++ b/lib/heroku/api/organizations_apps.rb
@@ -1,13 +1,13 @@
 module Heroku
   class API
-    def post_organizations_app_v3_dogwood(params={})
+    def post_organizations_app(params={})
       request(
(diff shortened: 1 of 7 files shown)


dickeyxxx:
lgtm

jsullivan:
I added one more commit to this PR after @dickeyxxx and I remembered that we still need to specify version 3 for any requests that return information about spaces and org apps. Might be worth a final once over before merging this in. 

Other than that, the platform api deploy that was blocking this PR is done so this can be merged whenever.

PR #1826, opened 2015-11-19
switch to osslsigncode to fix windows installs

dickeyxxx:


dickeyxxx:
@ransombriggs this seems to work

PR #1829, opened 2015-11-19
removed rendezvous specs since rendezvous is no longer used

dickeyxxx:


ransombriggs:
I went ahead and removed the one failing test in master, closing.

PR #1843, opened 2015-12-03
Fix Windows cryllic character from ENV issues

ransombriggs:
The underlying problem is that ENV variables with cryllic
characters come in as ASCII-8BIT rather than utf-8

* https://bugs.ruby-lang.org/issues/9715


ransombriggs:
@dickeyxxx please review

dickeyxxx:
lgtm

actually it doesn't, it's really ugly, but it should get the job done

PR #1849, opened 2015-12-08
Use system & exit on windows when non-ascii args

ransombriggs:


ransombriggs:
@dickeyxxx please review

dickeyxxx:
lgtm

PR #1860, opened 2015-12-18
Depend on ruby virtual package

adipascu:
This will fix the current issue on debian and
future proof the dependency.
libssl and libreadline are required by ruby.


dickeyxxx:
what issue does this fix exactly? I'm also not sure it will work for older ubuntu versions.

adi1133:
The issue is is a broken dependency in debian/testing. Debian currently lacks ruby 2.1 in favor of ruby 2.2

Instead of depending on a specific version i'd rather use the virtual package to prevent future dependency issues.

The ruby virtual package exists in all currently supported versions of ubuntu and debian, it also pulls in a ssl lib and the readline lib.

http://packages.ubuntu.com/search?keywords=ruby

https://packages.debian.org/search?keywords=ruby

isms:
@dickeyxxx Can confirm issue on Debian stretch:

```
$ sudo apt-get install heroku
Reading package lists... Done
Building dependency tree       
Reading state information... Done
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:

The following packages have unmet dependencies:
 heroku : Depends: ruby2.1 but it is not installable or
                   ruby2.0 but it is not installable or
                   libopenssl-ruby1.9.1 but it is not installable
          Depends: ruby2.1 but it is not installable or
                   ruby2.0 but it is not installable or
                   libreadline-ruby1.9.1 but it is not installable
          Depends: ruby2.1 but it is not installable or
                   ruby2.0 but it is not installable or
                   ruby1.9.1 but it is not installable
E: Unable to correct problems, you have held broken packages.
```
```
$ lsb_release -a
No LSB modules are available.
Distributor ID:	Debian
Description:	Debian GNU/Linux testing (stretch)
Release:	testing
Codename:	stretch
```

nikolai-b:
:+1: Looks like the ruby virtual package is in old Ubuntu and Debian repos
as https://github.com/heroku/heroku/pull/1860#issuecomment-166465350 said

PR #1876, opened 2016-01-14
delete unupdateable versions of v4

dickeyxxx:
will fix this: https://rollbar.com/heroku-cli/heroku-cli/items/584/


ransombriggs:
@dickeyxxx lgtm

PR #1880, opened 2016-01-19
remove duplicate command names in suggester

dickeyxxx:
because we actually have multiple commands that are the same
(ruby + node for example). We need to uniquely filter this list.

To repro: `heroku remote help`


rwz:
:+1:

PR #1884, opened 2016-01-24
Partition regions in Common Runtime and Private Spaces

ryanbrainard:
To match [what is shown in Devcenter](https://devcenter.heroku.com/articles/regions#view-the-list-of-available-regions), this partitions regions into separate `Common Runtime` and `Private Spaces` lists based on the new `private_capable` property introduced in v3.

cc: @dickeyxxx @jesperfj

### DIFF
--- a/lib/heroku/api/regions_v3.rb
+++ b/lib/heroku/api/regions_v3.rb
@@ -1,0 +1,14 @@
+module Heroku
+  class API
+    def get_regions_v3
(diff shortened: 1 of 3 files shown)


dickeyxxx:
closing since this went out in this https://github.com/heroku/heroku-apps/commit/6e6c0efba6fc6739d5571a0ed557d239305286ff

PR #1887, opened 2016-01-27
skip analytics on codeship

dickeyxxx:
codeship runs as a tty so it tries to prompt for the analytics

Fixes https://github.com/heroku/toolbelt/issues/145


ransombriggs:
lgtm

PR #1903, opened 2016-02-19
Clarify version for pg:upgrade

halorgium:
Often users will provide a `--version` flag, expecting to be able to choose the version to upgrade to. 
Currently, we upgrade to the latest stable version, so if another version is required, the `pg:copy` strategy should be used.

### DIFF
--- a/lib/heroku/command/pg.rb
+++ b/lib/heroku/command/pg.rb
@@ -463,7 +463,9 @@ class Heroku::Command::Pg < Heroku::Command::Base
 
   # pg:upgrade REPLICA
   #
-  # unfollow a database and upgrade it to the latest PostgreSQL version
+  # unfollow a database and upgrade it to the latest stable PostgreSQL version
+  #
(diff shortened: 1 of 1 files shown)


lib/heroku/command/pg.rb
```diff
@@ -463,7 +463,9 @@ def maintenance
 
   # pg:upgrade REPLICA
   #
-  # unfollow a database and upgrade it to the latest PostgreSQL version
+  # unfollow a database and upgrade it to the latest, stable PostgreSQL version
```
uhoh-itsmaciek: I'm no grammar expert, but 'latest' modifies 'stable' here, so I don't think you need a comma...

halorgium: I will fix it, I definitely am no grammar expert either.

neovintage:
sweet. thats awesome thanks @halorgium

wvengen:
:+1:

PR #1908, opened 2016-02-24
Remove code that lives in heroku-orgs

raulb:
Before this is merged, we still need to:

1. Cut a new release for github.com/heroku/heroku-orgs.
2. Merge this in: https://github.com/heroku/heroku-cli/pull/207.

@dickeyxxx Am I missing anything?


dickeyxxx:
typically we don't remove the ruby code, mostly because it can cause problems if any code has monkey patched this, but since in this case we want to remove some old command (right?) maybe we should

raulb:
@dickeyxxx maybe only for `heroku sharing` then.

dickeyxxx:
removing that would also require a changelog entry

PR #1913, opened 2016-03-14
Treat empty HEROKU_API_KEY the same as not set

ransombriggs:


ransombriggs:
@dickeyxxx review me?

dickeyxxx:
lgtm

PR #1915, opened 2016-03-18
Fix invalid byte sequence in format_with_bang

ransombriggs:


dickeyxxx:
the code is fine, but I'm not sure it's really necessary

PR #1924, opened 2016-04-05
re-enabled analytics

dickeyxxx:
this just shows a single message once notifying users them the analytics
will be enabled. This fixes problems with users using it in scripts and
is generally less annoying than it was before.


ransombriggs:
@dickeyxxx lgtm

Issue #1934, opened 2016-04-18
Updating... !    Heroku client internal error - Update hash signature mismatch


Madsvj:
Hi.

I'm running heroku on windows and while updating heroku

> heroku-cli: Updating... !    Heroku client internal error.

I get this error:
```
    Error:       Update hash signature mismatch (Heroku::Command::CommandFailed)
    Command:     heroku config --help
    Version:     heroku/toolbelt/3.42.40 (i386-mingw32) ruby/2.1.7
    Error ID:    2b30f2a7520245ec862ae4ff2b8fb490
```

The following is my **/.heroku/error.log**

```
Heroku client internal error.
Update hash signature mismatch
C:/Program Files (x86)/Heroku/lib/heroku/updater.rb:10:in `error'
C:/Program Files (x86)/Heroku/lib/heroku/updater.rb:132:in `block (2 levels) in update'
C:/Program Files (x86)/Heroku/ruby-2.1.7/lib/ruby/2.1.0/tmpdir.rb:88:in `mktmpdir'
C:/Program Files (x86)/Heroku/lib/heroku/updater.rb:121:in `block in update'
C:/Program Files (x86)/Heroku/lib/heroku/updater.rb:88:in `wait_for_lock'
C:/Program Files (x86)/Heroku/lib/heroku/updater.rb:117:in `update'
C:/Program Files (x86)/Heroku/lib/heroku/updater.rb:105:in `rescue in autoupdate'
C:/Program Files (x86)/Heroku/lib/heroku/updater.rb:101:in `autoupdate'
C:/Program Files (x86)/Heroku/lib/heroku/cli.rb:35:in `start'
C:/Program Files (x86)/Heroku/bin/heroku:29:in `<main>'
```

Googling suggests using a proxy server, but that didn't seem to work for me.
And I have no problems installing addons, or similar connections based problems.

** Windows version:

> OS Name:                   Microsoft Windows 7 Enterprise
> OS Version:                6.1.7601 Service Pack 1 Build 7601

ransombriggs:
@Madsvj Could you run the update command as follows to enable debugging for me?  
```
set HEROKU_DEBUG=1
heroku update
```

Madsvj:
Of course @ransombriggs 

Here is what it returns:

```
(venv) C:\Users\Bc0655\businesszone>set HEROKU_DEBUG=1

(venv) C:\Users\Bc0655\businesszone>heroku update
[...]

PR #1936, opened 2016-04-21
Fix bug where bad http status code are suppressed

ransombriggs:
@dickeyxxx review me?  I was digging into https://github.com/heroku/heroku/issues/1934#issuecomment-212772424 and it looks like bad status codes are not handled properly, this should now throw on non 200s after redirects.


dickeyxxx:
lgtm

PR #1941, opened 2016-05-01
Check and update the restore_from URL if it is from Dropbox

keiko713:
When users use a normal Dropbox URL for a dump file of `pg:backups restore` command, it fails with very ambiguous message:

````
$ heroku pg:backups restore 'https://www.dropbox.com/s/some-random-slug/rose.dump' HEROKU_POSTGRESQL_YELLOW_URL -a sushi
...
r014 ---restore---> HEROKU_POSTGRESQL_YELLOW
An error occurred and your backup did not finish.

Please run `heroku pg:backups info r014` for details.
```

```
$ heroku pg:backups info r013 -a sushi
...
=== Backup Logs
2016-05-01 18:42:53 +0000: pg_restore: [archiver] did not find magic string in file header
...
```

This is because that Dropbox URL actually gives you an HTML file, instead of the actual dump file. You can get the dump file [by specifying `dl=1` at the end of URL](https://www.dropbox.com/help/201), this PR is checking it and actually append it.
Alternatively, we could just raise the warning instead (that was my initial intention as you can see from the brunch name -- pg:backups command doesn't give you meaningful error messages in general), or we could do something at [transferatu](https://github.com/heroku/transferatu) or [htcat](https://github.com/htcat/htcat) level.

### DIFF
--- a/lib/heroku/command/pg_backups.rb
+++ b/lib/heroku/command/pg_backups.rb
@@ -405,7 +405,7 @@ EOF
 
     restore_url = nil
     if restore_from =~ %r{\Ahttps?://}
-      restore_url = restore_from
+      restore_url = check_dropboxurl(restore_from)
     else
       # assume we're restoring from a backup
       if restore_from =~ /::/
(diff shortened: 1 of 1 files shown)


dickeyxxx:
lgtm!

Issue #1901, opened 2016-02-18
pg:backups restore fails with a 0 exit code


jooshbzm:
A few times over the past month, one of our scheduled `pg:backups restore` jobs has failed silently and left the database in a bad state. From what I can tell, this command exits with a 0 exit code even when the restore fails.

Here's my output:

```
Running... 1.32GB
Running... 1.32GB
An error occurred and your backup did not finish.

Please run `heroku pg:backups info r303` for details.
```

ransombriggs:
@jooshbzm sorry this fell through the cracks, but a release has just been sent out so `heroku update` and verify that you have `heroku-toolbelt/3.43.1` and this should no longer occur

jooshbzm:
No problem @ransombriggs, I'll check this and confirm as soon as I get a chance. Thanks for fixing!

PR #1943, opened 2016-05-02
Exit non-zero code when pg:backups transfer fails

ransombriggs:
Fixes https://github.com/heroku/heroku/issues/1901


ransombriggs:
@dickeyxxx review me?

PR #1948, opened 2016-05-04
Change pgdiagnose endpoint

cyberdelia:
This change pgdiagnose, and I also took the occasion to make some stylistic changes.

cc/ @heroku/dod-core


dickeyxxx:
lgtm

PR #1958, opened 2016-05-25
retry v5 install

dickeyxxx:
not the most sophisticated retry logic, but should solve simple dns problems


ransombriggs:
@dickeyxxx lgtm

PR #1993, opened 2016-08-02
replace heroku-accounts warning with ruby plugin deprecation message

dickeyxxx:
We will wait until heroku-pg-extras is pushed before releasing this.

### DIFF
--- a/lib/heroku/cli.rb
+++ b/lib/heroku/cli.rb
@@ -30,7 +30,7 @@ class Heroku::CLI
     Heroku::Git.check_git_version
     Heroku::Command.load
     Heroku::Analytics.record(command)
-    warn_if_using_heroku_accounts
+    warn_if_ruby_plugin(command)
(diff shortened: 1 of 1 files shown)


dickeyxxx:
cc @ransombriggs @nsamsami @garyposter 

any tips on messaging here?

nsamsami:
The prompt seems a bit unclear. First line of warning that plugin is deprecated makes sense. Next line, not clear to me how to check if there's a new version or what Node.js Heroku CLI is (if I didn't know already.)

dickeyxxx:
One thing to keep in mind is only a handful of users will actually see this message and those that will are likely power users that have been using plugins probably for a few years (since no new ruby plugins have come out in a long time). So they'll likely understand even if it is somewhat vague.

Still I think this could be improved. I'm just not sure how exactly. It would be nice if we had a blog post when we launch the new CLI that we could link to here.

PR #1995, opened 2016-08-04
set commands to always run as ruby

dickeyxxx:
right now these are ignored because they are flagged as hidden in the
new CLI. Really though we don't want these hidden, so adding this
exception in allows us to show the commands in the new CLI and still
have the old one keep the old functionality.


dickeyxxx:
cc @ransombriggs

ransombriggs:
@dickeyxxx LGTM

PR #2004, opened 2016-08-29
Fix config path when XDG_CONFIG_HOME is set

smaximov:
Current version resolves `Heroku::Config.path` to `$XDG_CONFIG_HOME/config.json` instead of `$XDG_CONFIG_HOME/heroku/config.json` when `$XDG_CONFIG_HOME` is set, this patch fixes that.


smaximov:
Seems to be fixed by #1996.

PR #2011, opened 2016-09-07
Update rest-client gem because security vulnerabilities

ransombriggs:
@balazslaszlo I made two small changes for readability, but otherwise it looks good


dickeyxxx:
lgtm

PR #2013, opened 2016-09-08
Add appveyor test suite

ransombriggs:


ransombriggs:
@dickeyxxx not sure if you want this or not.  I am not sure how much value it adds given that it would not have caught the restclient update regression (see https://ci.appveyor.com/project/Heroku/heroku/build/1.0.5)

dickeyxxx:
lgtm

ransombriggs:
@dickeyxxx could you review again?  I removed `pry` and changed the `bundle install` in appveyor to be `--deployment` so that it fails if there is a Gem that cannot be installed from the `Gemfile.lock`.  So now if we have a PR like the one I merged yesterday, the build will fail with the following error, rather than hitting user machines.
```
Your Gemfile.lock is corrupt. The following gem is missing from the DEPENDENCIES section: 'ffi'
Command exited with code 20
```

dickeyxxx:
great! lgtm

PR #2017, opened 2016-09-23
Stop assuming domain for some private endpoints

cyberdelia:
Don't assume anything ``HEROKU_POSTGRESQL_HOST``, or think this is describing a specific subdomain.


dickeyxxx:
lgtm

PR #2031, opened 2016-10-28
Monkey patch URI to split urls with addressable

ransombriggs:
* Fixes bug where proxy url had underscore in it

@geemus could you review?  I decided to use addressable since we already had it as a gem dependency.


geemus:
👍  LGTM

